PR #333, opened 2025-12-14
Re-arrange SBOM export to provider level to avoid confusion on the build flow

yockgen:
#### Merge Checklist  

- [X] Ready to merge

#### Description 

Re-arrange SBOM export to provider level to avoid confusion on the build flow, this is to prepare for 100% accurate SBOM export for next step.

4 supporting provider been tested with dlstreamer enabled and verified running via QEMU


magerstam (approved):
Looks good

PR #342, opened 2026-01-07
The image inspect / compare ADR

magerstam:
#### Merge Checklist  

- [] The changes in the PR have been built and tested
- [] Ready to merge

#### Description 

This ADR provides an overview of the ability to inspect and image and generate a summary instead of relying on standard
Linux filesystem utilities in addition to be able to compare images between each other and generate a report. A similarity classification has been defined that allows users to get a better idea of how 'identical' two images are.

### DIFF
--- a/docs/architecture/adr-dep-analyzer.md
+++ b/docs/architecture/adr-dep-analyzer.md
@@ -1,0 +1,286 @@
+# ADR: Dependency Graph Analyzer Tool
+
+**Status**: Proposed  
+**Date**: 2026-01-08  
+**Updated**: N/A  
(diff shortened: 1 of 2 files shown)


yockgen:
It's super helpful for OS engineering teams when troubleshooting those classic complaints like 'the previous image worked fine, why doesn't the latest work?? I didn't change anything!'
Plus, Go has tons of libraries that make implementation easier (like github.com/diskfs/go-diskfs, libguestfs.org/guestfs, etc.) without needing to mount the raw image - which is great from a security standpoint.

PR #368, opened 2026-01-28
Enhanced comparison - introduced an equalityClass

magerstam:
#### Merge Checklist  

- [X] Ready to merge

#### Description 

- Removed the `Equal` flag to introduce levels of equality with images
- Semantically equal, Binary equal, Semantically equal (unverified)
- Introduced an optional flag --hash-images (default: false) which performs a SHA256 digest calculation on the provided images. This can take up to 20 seconds for 4GB images so made this optional for quicker comparison between images
- Updated unit tests
- The new logic with the Equality class helps compare images that *are* identical from a semantical perspective, but has differences that mutates each time an image is built, despite that it was generated from the same template (such as partition table UUIDs, file system UUIDs, etc.)
- 

(https://jira.devtools.intel.com/browse/ITEP-84987)

go-test

### DIFF
--- a/internal/image/imageinspect/compare.go
+++ b/internal/image/imageinspect/compare.go
@@ -43,8 +43,7 @@ type ImageDiff struct {
 
 // MetaDiff represents differences in image-level metadata.
 type MetaDiff struct {
-	File      *ValueDiff[string] `json:"file,omitempty"`
-	SizeBytes *ValueDiff[int64]  `json:"sizeBytes,omitempty"`
+	SizeBytes *ValueDiff[int64] `json:"sizeBytes,omitempty"`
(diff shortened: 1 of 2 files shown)


internal/image/imageinspect/compare.go
```diff
@@ -624,3 +701,224 @@ func appendFilesystemFieldChanges(dst []FieldChange, a, b FilesystemSummary) []F
 	}
 	return dst
 }
+
+func tallyDiffs(d ImageDiff) diffTally {
+	var t diffTally
+
+	if d.Image.File != nil {
```
arodage: refered this in above comment

PR #522, opened 2026-04-18
Add ADR for declarative live ISO installer

arodage:
#### Merge Checklist

- [x] Ready to merge

#### Description

This PR adds an Architecture Decision Record (ADR) proposing to extend the existing Live ISO installer into a **declarative provisioning engine** - eliminating the need for a separate interim OS provisioning environment.

The ADR addresses the following gaps in the current live installer:

1. **Automated disk selection** - policy-based disk discovery (largest, first, by-id, by-serial) instead of hardcoded `disk.path`
2. **Full Disk Encryption (FDE)** - LUKS2 provisioning with optional TPM enrollment and recovery key generation
3. **SELinux enforcement** - declarative mode/policy configuration with first-boot or install-time relabeling
4. **Network configuration** - template-driven netplan/systemd-networkd/NetworkManager config generation
5. **Install manifest separation** - decouple artifact production from provisioning logic

The proposal recommends a phased, incremental delivery (5 milestones) where M1, M3, M4 can be developed in parallel.

**Key design principles:**
- Single provisioning path for all use cases (BKC, Robotics, Edge)
- Fully declarative - all behavior driven by image template YAML
- Backward compatible - existing templates work without modification
- Clear separation: installer (low-level provisioning) vs cloud-init (high-level customization)

None.

- Documentation review only, no code changes


docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,610 @@
...
+---
+
+## Context
+
+### Problem Statement
+
+The OS Image Composer (ICT) generates fully qualified images with all required
+packages and tools, but is not a complete edge node provisioning and
+installation solution. Several capabilities required for production edge
+deployments are not available through the current Live ISO installer:
```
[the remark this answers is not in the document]
arodage: ICT is new name

docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,611 @@
...
+
+1. **Manual disk selection** - The unattended installer requires a hardcoded
+   `disk.path` (e.g., `/dev/sda`) in the template. There is no automatic disk
+   discovery or policy-based selection.
+
+2. **No Full Disk Encryption** - The boot parameter template contains
+   `{{.LuksUUID}}` and `{{.EncryptionBootUUID}}` placeholders, but they are
+   replaced with empty strings. No LUKS provisioning logic exists.
+
+3. **No SELinux automation** - The boot parameter template contains a
```
magerstam, line 42: We should (and I think we already do) install SELinux required packages, having a script that configures the SELinux policy (disabled, permissive, enforced). The latter may not be there as of today. We need to have a mechansim (perhaps similar to how we add secure boot signing files) reference in the template to link to a SELiunx policy file that can be provided by the customer.

docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,611 @@
...
+   `{{.SELinux}}` placeholder, also replaced with an empty string. SELinux
+   packages can be installed manually, but there is no automated mode
+   configuration, policy selection, or filesystem relabeling.
+
+4. **No declarative network configuration** - Network setup for the installed
```
magerstam, line 47: Enabling support for static addresses is inherently tricky here by the time you define the template. We should have at least two things beyond standard DHCP configuration of the NIC
- Selection of NIC (configure either all, or the one(s) with a 'link up' state)
- WiFi configuration (future)
- Proxy configuration (manual, and later we could see if we can support wdap and automatic proxy configuration through DHCP)

docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,611 @@
...
+   OS is not part of the template schema. Post-install networking relies on
+   cloud-init or manual `configurations` commands.
```
magerstam, line 49: Cloud init file would have to be supported to be injected by the customer at the time the image is built as we cannot assume you would have network connectivity once the ISO installation completes.

docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,611 @@
...
+### Existing Infrastructure
+
+The current codebase provides a strong foundation:
+
+| Capability | Status | Location |
+|---|---|---|
+| Disk enumeration via `lsblk` | Implemented (attended TUI only) | `imagedisc.SystemBlockDevices()` |
+| ISO media exclusion | Implemented | `imagedisc.isReadOnlyISO()` |
+| GPT/MBR partition creation | Implemented | `imagedisc.DiskPartitionsCreate()` |
+| dm-verity (hash partition + root hash) | Implemented | `imageos.prepareVeritySetup()` |
```
magerstam, line 81: I thought this has not yet been implemented in the live-installer flow?

docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,611 @@
...
+- Filesystem creation and formatting
+- Full Disk Encryption (LUKS2 provisioning)
+- dm-verity setup (hash partition, root hash injection)
+- SELinux base configuration (mode, policy, relabeling)
+- Bootloader installation (GRUB2, systemd-boot, UKI)
+- Root filesystem deployment
+
+### Cloud-init (post-first-boot)
+
+Handles customer customization:
```
magerstam, line 148: Need a mechanism through which the cloud-init file can be provided to the final image from the composer

docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,611 @@
...
+---
+
+## Phased Implementation
+
+### Phase 1: Automated Disk Selection
+
+**Problem**: The unattended installer requires `disk.path: /dev/sdX` to be
+hardcoded in the template - a value that varies across hardware.
+
+**Solution**: Add a `selectionPolicy` field to `DiskConfig` that allows the
```
magerstam, line 188: We can start with simple hierarchy
oneOf LargestDisc, FastestDisc
optional: excludeRemovableMedia

docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,611 @@
...
+#### Template Schema
+
+```yaml
+disk:
+  # Explicit path (existing behavior, still supported):
+  # path: /dev/sda
+
+  # New: policy-based selection for unattended installs
+  selectionPolicy:
+    strategy: largest           # largest | first | by-id | by-serial | by-model
```
magerstam: by-id, by-serial is going to be hard to scale. We have one template that can be installed on my systems so the id or serial is going to be a massive set of IDs

docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,611 @@
...
+When `disk.path` is set, the policy is ignored (backward compatible).
+
+#### Approach
+
+- Add a `DiskSelectionPolicy` struct to the config and a new disk selection
+  module in `internal/image/imagedisc/`
+- Reuse the existing `SystemBlockDevices()` enumeration, extending the
+  `lsblk` query to include `SERIAL`, `TRAN` (transport), and `RM` (removable)
+  fields
+- Implement strategy-based selection: `largest`, `first`, `by-id`,
```
magerstam: I would defer by-id and by-serial for now...

docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,611 @@
...
+  5. Optionally enroll TPM2 via `systemd-cryptenroll`
+  6. Optionally generate a recovery key saved to the ESP
+- Populate the existing `{{.LuksUUID}}` boot parameter placeholder with
+  `rd.luks.uuid=<uuid>` instead of the current empty string
+- Add `systemd-cryptenroll` to the shell command allowlist (`cryptsetup`
+  is already present)
+
+#### Security Considerations
+
+- Passphrase for non-TPM scenarios must be provided via template or secure
```
magerstam, line 394: The idea here is to configure automatic mounting with the passphrase stored in the TPM, not manually having to unlock with user interaction is that correct, or are we suggesting two modes
- interactive
- TPM PCR based automatic mounting
-

docs/architecture/adr-declarative-installer.md
```diff
@@ -0,0 +1,611 @@
...
+
+```yaml
+systemConfig:
+  network:
+    backend: netplan            # netplan | networkmanager | systemd-networkd
+    interfaces:
+      - name: eth0
+        dhcp4: true
+      - name: eth1
+        addresses:
```
magerstam: I think this is goign to be hard to scale for static assignment.

docs/architecture/adr-declarative-installer.md
arodage: Good point about customer-provided policy files. This comment is on the Problem Statement section. The proposed solution is in Phase 3: SELinux Enforcement below, which covers mode configuration (enforcing/permissive/disabled) and policy type selection (targeted/mls/minimum).

The SELinux config block (mode/policy/relabel) handles the base setup, turning SELinux on, selecting the policy type, and triggering relabeling. A .pp policy file can't control those; it only adds custom rules on top of the base policy.

We'll add a policyFiles field to support customer-provided modules alongside the config block:
```
selinux:
  mode: enforcing
  policy: targeted
  relabel: first-boot
  policyFiles:              # optional: custom policy modules
    - /path/to/custom.pp
```

See updated  Phase 3: SELinux Enforcement

magerstam (approved):
Looks good

PR #538, opened 2026-04-30
Fix/dot file naming bug

chrngc:
#### Merge Checklist  

- [x] Ready to merge

#### Description 

libpcre2-16-0 and libpcre2-32-0 was processed incorrectly previously, there were recognized as libpcre2 due to a word processing logic error. The method is flawed and this commit tries to minimize the use of the extractBasePackageNameFromFile to extract the canonical name, only use it as a fallback method when the canonical name isn't available. 

RequiresPkgNames is introduced as a new item in the PackageInfo struct, this item contains canonical name of required package. It is only used for GenerateDot for now, but could be useful for future feature implementation.
Before:
before
After:
after

### DIFF
--- a/internal/ospackage/ospackage.go
+++ b/internal/ospackage/ospackage.go
@@ -2,20 +2,21 @@ package ospackage
 
 // PackageInfo holds everything you need to fetch + verify one artifact.
 type PackageInfo struct {
-	Name        string // e.g. file name "abseil-cpp.rpm"
-	Type        string // e.g. "rpm", "deb", "apk"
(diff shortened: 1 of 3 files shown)


arodage:
@chrngc  it will be good if you can add some svg o/p to visualize the difference (before and after)

chrngc:
> @chrngc it will be good if you can add some svg o/p to visualize the difference (before and after)

Updated the description with before vs after image attached

arodage (approved):
LGTM

PR #545, opened 2026-05-04
feature: first-pass unattended ISO installer with policy-based target disk selection

magerstam:
#### Merge Checklist  

- [X] Ready to merge

#### Description 

- Adds support for building and booting a new Ubuntu 24 unattended installer ISO, including a dedicated image template, an unattended initrd default config, and a helper build script for local build and QEMU validation flows.
- Extends template/config handling to support policy-based install disk selection when `disk.path` is not explicitly set. This includes schema, merge, and config model updates for `selectionPolicy`.
- Updates the live installer flow to resolve the install target disk at runtime instead of requiring a fixed disk path in the template, and logs the selected disk for easier debugging.
- Hardens installer boot scripts for both attended and unattended flows by improving shell safety, quoting, retry handling, and ISO mount detection. The scripts now support both expected ISO labels and a `/dev/sr0` fallback.
- Improves disk discovery and selection logic by consuming richer `lsblk` metadata, filtering to real installable disks, excluding unsafe candidates such as device-mapper nodes, and making the "fastest" heuristic explicit and easier to reason about.
- Adds broader regression coverage around disk discovery, loop-device handling, config merge behavior, Ubuntu provider behavior, shell execution, file helpers, and bootloader configuration generation.
- Updates Ubuntu unattended initrd behavior to mask background `apt-daily` services and timers in the live environment so package refresh remains under installer control.
- Refreshes supporting repository assets included in the branch delta: CI workflow action pin updates, coverage threshold ratchet, template cleanup, a new Debian 13 desktop virtualization ISO template, and branch hygiene cleanup for generated local artifacts.

- Fixes unattended ISO installs failing when no explicit target disk path is provided by resolving the install disk from policy at install time.
- Fixes unsafe automatic disk selection by excluding non-disk and device-mapper entries from install candidates and supporting removable-disk exclusion.
- Fixes opaque "fastest disk" behavior by replacing arbitrary scoring with a clearer ordered comparison based on transport, rotational characteristics, and size tie-breaks.
- Fixes intermittent ISO startup/install issues caused by fragile ISO mount assumptions by making mount retry logic more robust across label and device fallback paths.
- Fixes CI and test failures caused by always prepending `sudo`, even when commands are already running as root.
- Fixes imageboot test instability by restoring global state correctly and asserting generated GRUB configuration more deterministically.
- Fixes live-installer UI regressions caused by stricter disk filtering by updating `lsblk` fixtures and making `DiskView` safe when partially initialized.
- Fixes intermittent Ubuntu unattended installer failures caused by apt lock contention in the live ISO environment by masking background `apt-daily` services/timers while still performing the installer’s explicit `apt update`.
- Fixes branch hygiene issues by ignoring root-level generated binaries/logs and removing accidentally included local editor artifacts.


internal/image/imagedisc/imagedisc.go
```diff
@@ -925,6 +963,284 @@ func SystemBlockDevices() (systemDevices []SystemBlockDevice, err error) {
...
+		eligible = append(eligible, dev)
+	}
+
+	if len(eligible) == 0 {
+		return "", fmt.Errorf("no supported disks found after applying selection policy")
+	}
+
+	strategy := strings.TrimSpace(strings.ToLower(diskConfig.SelectionPolicy.Strategy))
+	if strategy == "" {
+		strategy = DiskSelectStrategyLargest
```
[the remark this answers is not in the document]
magerstam: This has been addressed, Unattended installs no longer default an empty disk.path to largest disk. Explicit selectionPolicy.strategy is required to be defined.

config/general/isolinux/unattendedinstaller
```diff
@@ -8,7 +8,24 @@ CACHE_REPO=$ISO_ROOT/cache-repo
 
 # Mounting the ISO root for the installer.
 mkdir -p $ISO_ROOT
-LABEL=OIC_CDROM
+LABELS=(ICT_CDROM OIC_CDROM)
```
arodage, line 13: we dont need OIC_CDROM

config/general/isolinux/attendedinstaller
```diff
@@ -8,7 +8,24 @@ CACHE_REPO=$ISO_ROOT/cache-repo
 
 # Mounting the ISO root for the installer.
 mkdir -p $ISO_ROOT
-LABEL=OIC_CDROM
+LABELS=(ICT_CDROM OIC_CDROM)
```
arodage, line 13: We do not need OIC_CDROM

internal/image/imagedisc/imagedisc.go
```diff
@@ -876,7 +898,8 @@ func SystemBlockDevices() (systemDevices []SystemBlockDevice, err error) {
 	)
 
 	blockDeviceMajorNumbers := []string{scsiDiskMajorNumber, mmcBlockMajorNumber, virtualDiskMajorNumber, blockExtendedMajorNumber}
-	cmd := fmt.Sprintf("lsblk -d --bytes -I %s -n --json --output NAME,SIZE,MODEL", strings.Join(blockDeviceMajorNumbers, ","))
+	cmd := fmt.Sprintf("lsblk -d --bytes -I %s -n --json --output NAME,SIZE,MODEL,SERIAL,TRAN,RM,ROTA",
+		strings.Join(blockDeviceMajorNumbers, ","))
```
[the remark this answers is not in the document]
magerstam: Added stricter disk selection filter to address this.

chrngc:
image
Stumbled across this error during unattended ISO installation on a hardware with nvme as the largest drive. The script is trying to mount nvme0n1p1 but it's not created. Ran the attended installer that came along with the image and the ISO installation was able to complete without error.

internal/image/imagedisc/imagedisc.go
```diff
@@ -940,9 +1005,366 @@ func SystemBlockDevices() (systemDevices []SystemBlockDevice, err error) {
...
+	case DiskSelectStrategyFirst:
+		return eligible[0].DevicePath, nil
+	case DiskSelectStrategyLargest:
+		selected := eligible[0]
+		for _, dev := range eligible[1:] {
+			if dev.RawDiskSize > selected.RawDiskSize {
+				selected = dev
+			}
+		}
+		return selected.DevicePath, nil
```
chrngc, line 1155: Looks like this part will assign whatever comes first if I have two disks of the same size. I think it will produce a more consistent result if empty disk is always selected first when two disks have the same size, if both are empty and their size are exactly same, then choose the faster one. Only choose the first one that appears when both disks have the exact same setup and size

magerstam:
> image Stumbled across this error during unattended ISO installation on a hardware with nvme as the largest drive. The script is trying to mount nvme0n1p1 but it's not created. Ran the attended installer that came along with the image and the ISO installation was able to complete without error.

I need to debug this further to understand why this is really happening. Is it consistent or intermittent?

internal/image/imagedisc/imagedisc.go
magerstam: Yes, I agree with that and will address this. I will open a separate PR for this.

magerstam:
> > image Stumbled across this error during unattended ISO installation on a hardware with nvme as the largest drive. The script is trying to mount nvme0n1p1 but it's not created. Ran the attended installer that came along with the image and the ISO installation was able to complete without error.
> 
> I need to debug this further to understand why this is really happening. Is it consistent or intermittent?

I have noticed several race conditions, that appear to be surfacing in the unattended installer which requires some general hardening. I have opened a separate PR to address these more broadly here. #551 It would be good to get that PR reviewed, and merged in first. I can then rebase this PR on top of that as it is starting to grow a bit in scope here.

samueltaripin (commented):
i notice that it dint select a true empty disk. in my scenario,
1. disk list:
- usb (iso installer)
- nvme0n1 (ubuntu installed)
- nvme1n1 (empty - wiped with cmd "wipefs -a /dev/nvme1n1")

2. unattended ISO installer wipe ubuntu on nvme0n1 instead of selecting nvme1n1

magerstam:
> i notice that it dint select a true empty disk. in my scenario,
> 
> 1. disk list:
> 
> * usb (iso installer)
> * nvme0n1 (ubuntu installed)
> * nvme1n1 (empty - wiped with cmd "wipefs -a /dev/nvme1n1")
> 
> 2. unattended ISO installer wipe ubuntu on nvme0n1 instead of selecting nvme1n1

Correct, it doesn't require the disk to be empty. I think this feedback is good. It might be a lot cleaner to do the following (1) Remove 'largest-free' option in strategy. (2) Add another boolean parameter 'requireEmpty' which defaults to true. If you want to consider non-empty disks as well, override that flag, i.e.
```
"requireEmpty": {
      "type": "boolean",
      "description": "Require the selected disk to be empty before installation. When false, installation may overwrite existing data on the selected disk.",
      "default": true
    }
``` I will open a separate PR to add this functionality.

File: internal/image/imagedisc/imagedisc_test.go [hunk 6]
```diff
@@ -1263,6 +1280,244 @@ func TestSystemBlockDevices(t *testing.T) {
 	}
 }
 
+func boolPtr(v bool) *bool {
+	return &v
+}
+
+func TestResolveInstallDiskPath(t *testing.T) {
```
arodage, line 1287: Should we add a test for "zero eligible disks"?

internal/config/schema/os-image-template.schema.json
```diff
@@ -83,6 +83,23 @@
...
+              "description": "Disk selection strategy",
+              "enum": ["first", "largest", "fastest", "largest-free"]
+            },
+            "excludeRemovable": {
+              "type": "boolean",
+              "description": "Exclude removable or externally attached disks from selection using RM, transport, hotplug, udev, and sysfs heuristics",
+              "default": true
+            }
+          },
+          "additionalProperties": false
```
arodage, line 101: Better to add strategy as required attribute:
`"required": ["strategy"]`

arodage (approved):

internal/config/schema/os-image-template.schema.json
magerstam: Let's discuss this one separately, then we would need this section to be present in all of the templates... not just for unattended installation

internal/image/imagedisc/imagedisc_test.go
magerstam: I will add as i'm opening the second PR for improving the disk policy selection

PR #582, opened 2026-05-26
docs(architecture): add ADR for template extends

arodage:
#### Merge Checklist  

- [ ] Ready to merge

#### Description 

Adds an Architecture Decision Record (ADR) proposing `extends` support for image templates.

**Problem:** Users who need to customize existing templates must copy the entire template and modify it. When the original template is updated, every copy must be manually re-synchronized.

**Proposal:** Add an optional `extends` field to user templates that enables linear inheritance chains. Each template may extend at most one parent (single inheritance, no diamond problem). The merge is applied iteratively using the existing `MergeConfigurations()` as a fold operation, producing deterministic results at any chain depth.

Key design decisions:
- Reuses existing `MergeConfigurations()` applied iteratively (OS defaults -> root -> level 1 -> ... -> leaf)
- Single extends per template (no multiple inheritance) keeps merge deterministic
- Cycle detection via visited-set check
- Validation for target mismatches across the chain
- Depth warning at 4+ levels (soft limit, no hard cap)
- New `resolve` subcommand to inspect the merged result
- Local file paths only (no remote URL support initially)

None. This is a documentation-only change.

Documentation-only change. No code changes to test.

### DIFF
--- a/docs/architecture/adr-template-extends.md
+++ b/docs/architecture/adr-template-extends.md
@@ -76,16 +76,18 @@ systemConfig:
 ```mermaid
 flowchart TD
     A[Load child template] --> B{Has extends?}
-    B -- No --> G[Load OS default template]
+
+    B -- No --> G1[Load OS default template]
+    G1 --> I["Merge: MergeConfigurations(child, default) → final"]
+
     B -- Yes --> C[Resolve path relative to child location]
     C --> D[Load parent template]
     D --> E{Parent has extends?}
     E -- Yes --> ERR1[❌ Reject: single-level only]
     E -- No --> F{Target matches child?}
     F -- No --> ERR2["❌ Reject: target mismatch (OS/dist/arch/imageType)"]
-    F -- Yes --> G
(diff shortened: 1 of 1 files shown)


docs/architecture/adr-template-extends.md
```diff
@@ -0,0 +1,198 @@
...
+**Date**: 2026-05-26  
+**Updated**: 2026-05-30  
+**Authors**: ICT Team  
+**Technical Area**: Template Configuration / Merge System
+
+---
+
+## Summary
+
+Add an optional `extends` field to user templates, enabling linear inheritance chains. Each template may extend at most one parent template. The merge is applied iteratively from the root of the chain through each level, producing a deterministic result. This allows users to maintain minimal delta templates that automatically inherit updates from their parent templates.
```
magerstam, line 13: I thought we had agreed to allow for this to be arbitrarily extended in a linear fashion? I see that this is how it has been described further down, perhaps just fix this typo

magerstam (approved):
This looks good to me Alpesh

PR #583, opened 2026-05-27
feat: add graceful loop device cleanup on Ctrl+C interrupt

rpandya28:
#### Merge Checklist  

- [ ] The changes in the PR have been built and tested
- [ ] Ready to merge


internal/image/rawmaker/rawmaker.go
```diff
@@ -143,6 +154,27 @@ func (rawMaker *RawMaker) BuildRawImage() error {
 		}
 	}()
 
+	// Goroutine waits for either an OS signal or normal completion.
+	// Using select + a dedicated done channel avoids closing sigChan and
+	// makes the two paths explicit: real interrupt vs. function returned.
+	go func() {
```
arodage, line 166: This goroutine calls os.Exit(1), the entire process terminates immediately. os.Exit does not run any deferred functions, not the defer in BuildRawImage at rawmaker.go:145, not defers in the provider, not defers in main(). So chroot mounts, temp directories, and any other resources managed higher in the stack are leaked.

The solution is to not call `os.Exit` from library code and instead move signal handling up to the cmd level where it can orchestrate full cleanup.

arodage, line 166: Race condition: imagine a `Ctrl+C` arrives while the build is finishing normally. The signal gets queued in `sigChan`. Then the `defer` runs, it calls `signal.Stop` and `close(done)`. At this point the goroutine's `select` sees two ready cases: the queued signal in `sigChan` AND the closed `done` channel. Go picks one at random. If it picks the signal case, both the goroutine and the defer end up calling `LoopSetupDelete` on the same loop device at the same time. Wrap the cleanup in a `sync.Once` so only one of them actually runs it.

internal/image/imagedisc/loopdev.go
```diff
@@ -60,11 +60,105 @@ func (loopDev *LoopDev) LoopSetupDelete(loopDevPath string) error {
...
+	// Find and unmount all mounted partitions
+	return loopDev.findAndUnmount(result)
+}
+
+// findAndUnmount recursively finds mounted partitions and unmounts them
+func (loopDev *LoopDev) findAndUnmount(data interface{}) error {
+	switch v := data.(type) {
+	case map[string]interface{}:
+		// Check if this entry has a mountpoint
+		if mountPoint, ok := v["mountpoint"].(string); ok && mountPoint != "" && mountPoint != "null" {
```
arodage: After json.Unmarshal, a JSON null becomes Go nil, not the string "null". The .(string) type assertion already filters out nil values, so this comparison `mountPoint != "null"` is dead code.

arodage:
Need to add unit test for new functions. At minimum for  `findAndUnmount` and `findAndDisableSwap`

arodage:
already taken care by another PR

PR #676, opened 2026-06-08
fix chroot stale cache not clean up

samueltaripin:
#### Merge Checklist  

- [x] Ready to merge

#### Description 

This PR fixes stale DEB package cache cleanup by ensuring cached .deb files are removed not only from the top-level cache directory but also from nested subdirectories used during image builds (e.g., chroot/initrd caches). This helps prevent stale packages from being indexed and shadowing the intended versions in subsequent runs.

Changes:

Update DEB cache cleanup to recursively remove .deb files under the cache directory and gracefully no-op when the cache directory is missing.
Expand unit tests to cover nested subdirectory cleanup and the missing-directory no-op behavior.


arodage (approved):
Looks good.

PR #704, opened 2026-06-22
fix(unattended-installer): harden disk selection startup timing and logging

elvin03:
#### Merge Checklist  

- [x] Ready to merge

#### Description 

Improve unattended install reliability on multi-console early-boot flows and USB hotplug races.

- add persistent unattended log file output via tee to /tmp/unattended-installer.log
- restore console file descriptors before opening failure shell so prompt is interactive
- add single-instance lock in unattended launcher to prevent duplicate concurrent runs
- log disk selection policy and discovered candidates at info level for easier triage
- add udevadm settle before disk discovery in ResolveInstallDiskPath
- allow udevadm in shell command allowlist so settle can execute
- keep detailed candidate evaluation output when no eligible disk matches policy

This addresses cases where installer starts too early, only sees NVMe, and misses later-enumerated USB targets.

N/A

Before installing
sdb_0

during installing
sdb_1

after installing
sdb_2


polmoorx:
I verified this PR locally, With the template settings requireempty = true and excludedremovable = false, I tested a scenario where the ISO installer was loaded on the SATA disk (sda) and sdb was an empty disk. In this case, the installer automatically selected sdb as the target disk and successfully installed the OS on it, which is the expected behavior.

samueltaripin (approved):

PR #714, opened 2026-06-26
Add baseline and overlayPolicy to template schema

yktyagi:
Defines the template contract for overlay mode and wires it through config merge so a user-provided baseline survives into the build.

Config (internal/config/config.go):
  - Baseline{Mode, Source, OverlayPolicy}, BaselineSource{Path, Format}, OverlayPolicy{PackageOperation, ConflictPolicy, AllowDowngrade, KernelCmdline, AllowRemoval}
  - validateBaseline: nil baseline behaves as create; create rejects source/overlayPolicy; overlay requires a non-empty local source path, rejects any URL scheme, and accepts only format=raw
  - OverlayPolicy.validate defaults and checks packageOperation (additive-only) and conflictPolicy (fail|allow-explicit)
  - IsOverlayMode helper
  - AllowRemoval is a non-YAML field pinned false: overlay is additive-only in v1, so the schema never surfaces the knob

Merge (internal/config/merge.go):
  - MergeConfigurations now carries the user's baseline over the default template; without this an overlay baseline was silently dropped and the build fell back to create mode

Schema (os-image-template.schema.json):
  - $defs Baseline/BaselineSource/OverlayPolicy with enums and defaults
  - additionalProperties:false rejects allowRemoval and remote URL sources at the schema layer

Tests (baseline_test.go, merge_test.go):
  - table-driven TestValidateBaseline (defaults, overlay-requires-source, create-rejects-source, format=raw only, omitted baseline = create, plus enum and URL-scheme rejection incl. file://), TestIsOverlayMode, schema-layer accept/reject tests, and TestMergeConfigurationsBaseline

#### Merge Checklist  

- [ ] Ready to merge

### DIFF
--- a/internal/config/baseline_test.go
+++ b/internal/config/baseline_test.go
@@ -1,0 +1,331 @@
+package config
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/open-edge-platform/image-composer-tool/internal/config/validate"
+)
+
+func TestValidateBaseline(t *testing.T) {
+	tests := []struct {
+		name          string
+		baseline      *Baseline
+		overlayPolicy *OverlayPolicy
+		wantErr       string
+		wantNoErr     bool
+	}{
+		{
+			name:      "nil baseline is allowed (equivalent to create)",
+			baseline:  nil,
+			wantNoErr: true,
+		},
+		{
+			name:      "explicit create with no source is allowed",
+			baseline:  &Baseline{Mode: BaselineModeCreate},
+			wantNoErr: true,
+		},
+		{
+			name:      "empty mode defaults to create",
+			baseline:  &Baseline{},
+			wantNoErr: true,
+		},
+		{
+			name:     "create rejects source",
+			baseline: &Baseline{Mode: BaselineModeCreate, Source: &BaselineSource{Path: "/tmp/x.raw"}},
+			wantErr:  "source must not be set",
+		},
+		{
+			name:          "create rejects overlayPolicy",
+			baseline:      &Baseline{Mode: BaselineModeCreate},
+			overlayPolicy: &OverlayPolicy{},
+			wantErr:       "overlayPolicy must not be set",
+		},
+		{
+			name:          "overlayPolicy without baseline (default create) is rejected",
+			baseline:      nil,
+			overlayPolicy: &OverlayPolicy{},
+			wantErr:       "overlayPolicy must not be set",
+		},
+		{
+			name:     "overlay requires source",
+			baseline: &Baseline{Mode: BaselineModeOverlay},
+			wantErr:  "source is required",
+		},
+		{
+			name:      "overlay with valid local source path passes",
+			baseline:  &Baseline{Mode: BaselineModeOverlay, Source: &BaselineSource{Path: "/tmp/u.raw"}},
+			wantNoErr: true,
+		},
+		{
+			name: "overlay accepts https URL source",
+			baseline: &Baseline{
+				Mode:   BaselineModeOverlay,
+				Source: &BaselineSource{URL: "https://example.com/u.raw"},
+			},
+			wantNoErr: true,
+		},
+		{
+			name: "overlay accepts http URL source",
+			baseline: &Baseline{
+				Mode:   BaselineModeOverlay,
+				Source: &BaselineSource{URL: "http://example.com/u.raw"},
+			},
+			wantNoErr: true,
+		},
+		{
+			name: "overlay rejects URL written into path field",
+			baseline: &Baseline{
+				Mode:   BaselineModeOverlay,
+				Source: &BaselineSource{Path: "https://example.com/u.raw"},
+			},
+			wantErr: "use baseline.source.url for remote images",
+		},
+		{
+			name: "overlay rejects single-slash scheme in path field",
+			baseline: &Baseline{
+				Mode:   BaselineModeOverlay,
+				Source: &BaselineSource{Path: "file:/tmp/u.raw"},
+			},
+			wantErr: "use baseline.source.url for remote images",
+		},
+		{
+			name: "overlay rejects non-http URL scheme",
+			baseline: &Baseline{
+				Mode:   BaselineModeOverlay,
+				Source: &BaselineSource{URL: "file:///tmp/u.raw"},
+			},
+			wantErr: "must use http or https",
+		},
+		{
+			name: "overlay rejects source with neither path nor url",
+			baseline: &Baseline{
+				Mode:   BaselineModeOverlay,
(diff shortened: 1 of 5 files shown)


internal/config/schema/os-image-template.schema.json
```diff
@@ -440,6 +440,68 @@
...
+      "description": "Baseline image mode selector. Default mode is 'create' (build from scratch). 'overlay' layers packages onto an existing baseline image.",
+      "properties": {
+        "mode": {
+          "type": "string",
+          "description": "Baseline assembly mode",
+          "enum": ["create", "overlay"],
+          "default": "create"
+        },
+        "source": { "$ref": "#/$defs/BaselineSource" },
+        "overlayPolicy": { "$ref": "#/$defs/OverlayPolicy" }
```
arodage: As per the ADR examples and schema recommendation, show `overlayPolicy` as a top-level (peer to `Baseline`)
Here we have it under Baseline.
Either approach works but ADR and implementation should be aligned.

internal/config/schema/os-image-template.schema.json
```diff
@@ -440,6 +440,68 @@
...
+      },
+      "additionalProperties": false
+    },
+    "BaselineSource": {
```
arodage, line 472: Schema does not enforce mode/source coupling.
The schema defines baseline.mode and baseline.source but does not include if/then rules requiring source for overlay and rejecting source for create. That logic exists only in Go validation.

Refer following section in ADR:
  "allOf": [
    {
      "if": {
        "properties": {
          "mode": { "const": "overlay" }
        }
      },
      "then": {
        "required": ["source"]
      }
    },
    {
      "if": {
        "properties": {
          "mode": { "const": "create" }
        }
      },
      "then": {
        "not": {
          "required": ["source"]
        }
      }
    }
  ]

internal/config/baseline_test.go
```diff
@@ -0,0 +1,254 @@
...
+			baseline: &Baseline{Mode: BaselineModeOverlay},
+			wantErr:  "source is required",
+		},
+		{
+			name:      "overlay with valid local source path passes",
+			baseline:  &Baseline{Mode: BaselineModeOverlay, Source: &BaselineSource{Path: "/tmp/u.raw"}},
+			wantNoErr: true,
+		},
+		{
+			name: "overlay rejects URL-style source path",
```
arodage: Get confirmation on this.

arodage (approved):
LGTM. Can be closed once all comments are addressed.

internal/config/schema/os-image-template.schema.json
yktyagi: fixed

internal/config/schema/os-image-template.schema.json
yktyagi: fixed

internal/config/baseline_test.go
yktyagi: got confirmation. URL is supported through source.url field.

internal/config/config.go
```diff
@@ -92,11 +92,56 @@ type ProviderRepoConfigs struct {
...
+	PackageOperation string `yaml:"packageOperation,omitempty"`
+	ConflictPolicy   string `yaml:"conflictPolicy,omitempty"`
+	AllowDowngrade   bool   `yaml:"allowDowngrade,omitempty"`
+	KernelCmdline    string `yaml:"kernelCmdline,omitempty"`
+
+	// AllowRemoval gates whether preflight permits removing a baseline package.
+	// It is intentionally NOT a YAML field (the schema rejects it): overlay mode
+	// is additive-only in v1, so this is pinned false by effectiveOverlayPolicy.
+	// The gate reads it so the removal-blocking logic is exercised and a future
+	// release can lift the restriction by flipping a single default.
```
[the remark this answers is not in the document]
yktyagi: AllowRemoval is simply never assigned. It has yaml:"-" (excluded from YAML), and the schema's additionalProperties: false rejects any attempt to set allowRemoval in a template. With no code path writing to it, it always holds Go's zero value — false. There's no normalization function; the "pinned false" behavior just falls out of the zero value plus schema rejection.

internal/config/config.go
```diff
@@ -1060,3 +1109,77 @@ func (pr *PackageRepository) ValidatePackageRepository() error {
...
+	case BaselineModeOverlay:
+		if b.Source == nil {
+			return fmt.Errorf("baseline: source is required when mode is %q", BaselineModeOverlay)
+		}
+		if strings.TrimSpace(b.Source.Path) == "" {
+			return fmt.Errorf("baseline.source.path must be a non-empty local file path")
+		}
+		if strings.Contains(b.Source.Path, "://") {
+			return fmt.Errorf("baseline.source.path must be a local file path; URL schemes (http://, https://, file://, etc.) are not supported in v1")
+		}
```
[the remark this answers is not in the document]
yktyagi: fixed. Replaced the strings.Contains(path, "://") substring check with url.Parse, rejecting any non-empty scheme:
case path != "":
	// Reject any URI scheme (e.g. http:/, file:/path, file://...). Local
	// paths have no scheme; remote images belong in baseline.source.url.
	if parsed, err := url.Parse(path); err == nil && parsed.Scheme != "" {
		return fmt.Errorf("baseline.source.path must be a local file path; use baseline.source.url for remote images")
	}
This now catches single-slash forms like file:/path that :// missed, and the error message no longer over-claims — it just points users to baseline.source.url for remote images. Added a regression test (overlay rejects single-slash scheme in path field, Path: "file:/tmp/u.raw").

internal/config/schema/os-image-template.schema.json
```diff
@@ -440,6 +440,68 @@
...
+      "type": "object",
+      "description": "Baseline source. v1 supports a local RAW disk image only.",
+      "properties": {
+        "path": {
+          "type": "string",
+          "description": "Local filesystem path to the baseline RAW disk image",
+          "minLength": 1
+        },
```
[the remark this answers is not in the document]
yktyagi: Added the schema-layer guard. baseline.source.path now carries a pattern that rejects any value with a leading URI scheme, so a URL like https://example.com/u.raw is rejected at validation/load time, not just at runtime:


"path": {
  "type": "string",
  "description": "Local filesystem path to the baseline RAW disk image (no URI scheme; use 'url' for remote images)",
  "minLength": 1,
  "not": { "pattern": "^[A-Za-z][A-Za-z0-9+.-]*:" }
}
The pattern is RE2-safe (santhosh-tekuri/jsonschema uses Go's regexp), matching scheme: per RFC 3986 and rejecting it via not. Remote images belong in the dedicated baseline.source.url field. Added TestSchemaRejectsURLInPath to cover it.

PR #715, opened 2026-06-26
Simple script to extract rootfs from RAW

magerstam:
#### Merge Checklist  

- [X] Ready to merge

#### Description 

Simple shell script to extract the rootfs and compress it from a `RAW` image. The extracted
output allows for the rootfs to be mounted and used in WSL2 with the installed Windows linux kernel

### DIFF
--- a/scripts/extract-rootfs-from-raw.sh
+++ b/scripts/extract-rootfs-from-raw.sh
@@ -1,0 +1,134 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+IMAGE="${1:-image.raw}"
+OUT="${2:-ict-rootfs.tar}"
+PART_NUM="${3:-}"
+MOUNT_DIR="${MOUNT_DIR:-/tmp/ict-rootfs}"
+ADD_WSL_CONF="${ADD_WSL_CONF:-true}"
+
+usage() {
+  echo "Usage: $0 [image.raw] [output.tar|output.tar.gz] [partition_number]"
+  echo "Env vars: MOUNT_DIR=/tmp/ict-rootfs ADD_WSL_CONF=true|false"
+}
+
+require_cmd() {
+  if ! command -v "$1" >/dev/null 2>&1; then
+    echo "Missing required command: $1" >&2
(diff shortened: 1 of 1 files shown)


arodage:
It might be a good idea to add documentation for usage and add link in usage guide

scripts/extract-rootfs-from-raw.sh
```diff
@@ -0,0 +1,134 @@
...
+}
+
+require_cmd() {
+  if ! command -v "$1" >/dev/null 2>&1; then
+    echo "Missing required command: $1" >&2
+    exit 1
+  fi
+}
+
+for cmd in sudo fdisk losetup mount umount tar gzip mktemp; do
```
arodage, line 22: If a user passes -h, the script first validates that sudo, fdisk, losetup, … are installed before showing help. The help check should come before the require_cmd loop.

scripts/extract-rootfs-from-raw.sh
```diff
@@ -0,0 +1,134 @@
...
+  esac
+}
+
+echo "Creating rootfs archive: $OUT"
+create_archive "$OUT"
+
+echo "Done: $OUT"
+echo
+echo "On Windows, import with:"
+echo "wsl --import test-aiagent C:\\WSL\\test-aiagent C:\\path\\to\\$(basename "$OUT") --version 2"
```
arodage, line 134: hardcoded path

scripts/extract-rootfs-from-raw.sh
```diff
@@ -0,0 +1,134 @@
...
+fi
+
+LOOP=""
+ROOT_PART=""
+WSL_TMP_DIR=""
+
+cleanup() {
+  set +e
+  if mountpoint -q "$MOUNT_DIR" 2>/dev/null; then
+    sudo umount "$MOUNT_DIR" 2>/dev/null || true
```
arodage, line 49: do we need to delete MOUNT_DIR  after umount?

arodage:
Is this script added to git with executable permissions?

PR #716, opened 2026-06-29
feat(imageos): add rootfs/immutability safety checks for first-boot auto-expand

elvin03:
#### Merge Checklist  

- [x] Ready to merge

#### Description 

Add defensive checks for last-partition auto-expansion so it only activates for supported templates and runtime layouts.

Skip auto-expand asset injection when immutability is enabled
Skip auto-expand asset injection when the last partition is not mounted at /
Add runtime guard in first-boot script to ensure detected last partition matches root source
Extend tests to cover disabled flag, immutability-enabled, non-rootfs-last, and valid happy-path cases

N/A

Validated in RPL
Before boot to image
auto-expand-before

After boot to image
auto-expand-after


arodage:
Here is few issues we need to address:
1. GetEffectivePartitions() is dead code
The method is defined and tested but never called anywhere in the codebase. It should either be wired into the actual partition creation path (presumably in rawmaker) or removed. Currently the end: "0" is expected to be set directly in the template YAML rather than computed by this method.

2. Duplicate validation logic
validateFirstBootLastPartitionAutoExpandConfig() in [imageos.go](vscode-file://vscode-app/c:/Users/arodage/AppData/Local/Programs/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html) duplicates validateAutoExpandLastPartitionConstraints() in [validate.go](vscode-file://vscode-app/c:/Users/arodage/AppData/Local/Programs/Microsoft%20VS%20Code/4fe60c8b1c/resources/app/out/vs/code/electron-browser/workbench/workbench.html). Both check immutability + last-partition-is-rootfs. The imageos version should rely on the validation already done at template load time, or at minimum reference a shared function.

3. validateFirstBootLastPartitionAutoExpandConfig is called twice per build
Both injectFirstBootLastPartitionAutoExpandAdditionalFiles and setupFirstBootLastPartitionAutoExpand call it independently. Since both are called sequentially in updateRootfsConfig / updateImageConfig, this validation runs twice unnecessarily.

4. Shell script runtime guard vs. build-time guard redundancy
The shell script has its own runtime check (LAST_PART_DEV_REAL != ROOT_SRC_REAL), the validate.go has a build-time check, and imageos.go has yet another build-time check. Three layers of the same guard is excessive. The runtime guard makes sense (defensive), but the build-time duplication does not.

5. injectFirstBootLastPartitionAutoExpandAdditionalFiles mutates template state
It appends to template.SystemConfig.AdditionalFiles as a side effect before addImageAdditionalFiles runs. This coupling to call order is fragile — if someone reorders the calls in updateRootfsConfig, the additional files won't be copied. A comment explaining the ordering dependency would help, but a cleaner approach would be to handle the injection and copy as a single operation.

6. Schema doesn't restrict extendLastPartitionToFillDisk to raw imageType
The JSON schema adds the boolean at the disk level unconditionally. A template with imageType: iso and extendLastPartitionToFillDisk: true would pass schema validation but silently be ignored at build time. Consider adding a conditional schema rule or at least a warning.

7. DiskConfig field ordering is inconsistent
The new field ExtendLastPartitionToFillDisk is placed between SelectionPolicy and Artifacts, breaking the logical grouping. Fields like Partitions and PartitionTableType are more closely related. Minor, but the re-alignment of all other fields makes the diff noisier than needed.

8. Missing t.Parallel() in tests
Per project conventions, tests should use t.Parallel() where safe. The new test functions (TestSetupFirstBootLastPartitionAutoExpand*, TestGetEffectivePartitions, TestValidateImageTemplateJSON_AutoExpand*) don't use it. The imageos tests mutate global state (shell.Default, config.Global().ConfigDir) so parallel isn't safe there, but the config and validate tests could use it.

9. Validate tests don't follow table-driven pattern
The four TestValidateImageTemplateJSON_AutoExpand* tests repeat the same YAML-parse → JSON-marshal → validate pattern. Per project conventions, these should be a single table-driven test with t.Run.

elvin03:
> Here is few issues we need to address:
> 
> 1. GetEffectivePartitions() is dead code
>    The method is defined and tested but never called anywhere in the codebase. It should either be wired into the actual partition creation path (presumably in rawmaker) or removed. Currently the end: "0" is expected to be set directly in the template YAML rather than computed by this method.
> 2. Duplicate validation logic
>    validateFirstBootLastPartitionAutoExpandConfig() in imageos.go duplicates validateAutoExpandLastPartitionConstraints() in validate.go. Both check immutability + last-partition-is-rootfs. The imageos version should rely on the validation already done at template load time, or at minimum reference a shared function.
> 3. validateFirstBootLastPartitionAutoExpandConfig is called twice per build
>    Both injectFirstBootLastPartitionAutoExpandAdditionalFiles and setupFirstBootLastPartitionAutoExpand call it independently. Since both are called sequentially in updateRootfsConfig / updateImageConfig, this validation runs twice unnecessarily.
> 4. Shell script runtime guard vs. build-time guard redundancy
>    The shell script has its own runtime check (LAST_PART_DEV_REAL != ROOT_SRC_REAL), the validate.go has a build-time check, and imageos.go has yet another build-time check. Three layers of the same guard is excessive. The runtime guard makes sense (defensive), but the build-time duplication does not.
> 5. injectFirstBootLastPartitionAutoExpandAdditionalFiles mutates template state
>    It appends to template.SystemConfig.AdditionalFiles as a side effect before addImageAdditionalFiles runs. This coupling to call order is fragile — if someone reorders the calls in updateRootfsConfig, the additional files won't be copied. A comment explaining the ordering dependency would help, but a cleaner approach would be to handle the injection and copy as a single operation.
> 6. Schema doesn't restrict extendLastPartitionToFillDisk to raw imageType
>    The JSON schema adds the boolean at the disk level unconditionally. A template with imageType: iso and extendLastPartitionToFillDisk: true would pass schema validation but silently be ignored at build time. Consider adding a conditional schema rule or at least a warning.
> 7. DiskConfig field ordering is inconsistent
>    The new field ExtendLastPartitionToFillDisk is placed between SelectionPolicy and Artifacts, breaking the logical grouping. Fields like Partitions and PartitionTableType are more closely related. Minor, but the re-alignment of all other fields makes the diff noisier than needed.
> 8. Missing t.Parallel() in tests
>    Per project conventions, tests should use t.Parallel() where safe. The new test functions (TestSetupFirstBootLastPartitionAutoExpand*, TestGetEffectivePartitions, TestValidateImageTemplateJSON_AutoExpand*) don't use it. The imageos tests mutate global state (shell.Default, config.Global().ConfigDir) so parallel isn't safe there, but the config and validate tests could use it.
> 9. Validate tests don't follow table-driven pattern
>    The four TestValidateImageTemplateJSON_AutoExpand* tests repeat the same YAML-parse → JSON-marshal → validate pattern. Per project conventions, these should be a single table-driven test with t.Run.

Done rework

samueltaripin (approved):

PR #718, opened 2026-07-01
feat(config): add extends chain resolution algorithm

arodage:
#### Merge Checklist

- [ ] Ready to merge

#### Description

Implements the extends chain resolution algorithm from `docs/architecture/adr-template-extends.md`. Adds `resolveExtendsChain(templatePath string) ([]*ImageTemplate, error)` in `internal/config/merge.go`, which walks `extends` references from leaf to root, loads every template, and validates the chain.

Behavior:
- Resolves `extends` paths relative to each child template's directory
- Returns an ordered slice `[root, level1, ..., leaf]`
- **Cycle detection** via a visited-set keyed on canonicalized (symlink-resolved) paths; rejects `A -> B -> A` with `circular extends detected: a.yml -> b.yml -> a.yml`
- **Target match validation**: all templates must share OS/dist/arch/imageType; mismatch returns `extends target mismatch at level N: ...`
- **Path safety**: `filepath.Clean`, symlink rejection via `security.RejectSymlinks`, and rejection of paths escaping the child template's directory
- **Missing parent**: returns `extends path not found: <resolved-path>`
- **Depth warning**: logs a warning when the chain exceeds 4 extends levels (no hard rejection, per ADR)

Note: `resolveExtendsChain` is not yet wired into `LoadAndMergeTemplate`  that integration is intentionally deferred to a follow-up PR.

- Cycle detection is keyed on `filepath.EvalSymlinks`-canonicalized paths, closing a gap where a directory symlink could alias two textual paths to the same file and evade detection. Covered by `TestResolveExtendsChainCycleViaDirectorySymlink`.

None.

Unit tests in `internal/config/merge_test.go`:
- Single extends
- Multi-level chain (3+ levels)
- Cycle detection
- Directory-symlink cycle aliasing (hardening)
- Target mismatch
- Missing parent
- Path traversal attack
- Depth warning (captured via logger output)

All resolver tests and the full `go test ./internal/config/` suite pass; `go vet ./internal/config/...` is clean.

### DIFF
--- a/internal/config/merge.go
+++ b/internal/config/merge.go
@@ -674,6 +674,13 @@ func resolveExtendsChain(templatePath string) ([]*ImageTemplate, error) {
 			return nil, fmt.Errorf("invalid extends path: %w", err)
 		}
 
+		// The lexical guard in resolveExtendsParentPath cannot see through directory
+		// symlinks inside the child directory. Re-check containment against the fully
+		// symlink-resolved paths so a symlink like child/link -> /outside cannot escape.
+		if err := verifyResolvedContainment(currentPath, parentPath); err != nil {
+			return nil, err
+		}
+
 		if cycleStart, exists := visited[canonicalPath(parentPath)]; exists {
 			cyclePaths := append([]string{}, displayPaths[cycleStart:]...)
 			cyclePaths = append(cyclePaths, parentPath)
@@ -725,6 +732,32 @@ func resolveExtendsParentPath(childTemplatePath, extendsRef string) (string, err
 	}
 
 	return parentPath, nil
+}
+
+// verifyResolvedContainment ensures that, after resolving all symlinks, the parent
+// template still lives within the child template's directory. This defends against
+// directory symlinks inside the child directory that the lexical check in
+// resolveExtendsParentPath cannot detect. Both paths must already exist.
+func verifyResolvedContainment(childTemplatePath, parentPath string) error {
+	resolvedChildDir, err := filepath.EvalSymlinks(filepath.Dir(childTemplatePath))
+	if err != nil {
+		return fmt.Errorf("failed to resolve child template directory: %w", err)
+	}
+
+	resolvedParent, err := filepath.EvalSymlinks(parentPath)
+	if err != nil {
+		return fmt.Errorf("failed to resolve extends path %s: %w", parentPath, err)
+	}
+
+	rel, err := filepath.Rel(resolvedChildDir, resolvedParent)
+	if err != nil {
+		return fmt.Errorf("failed to evaluate resolved extends path %s: %w", parentPath, err)
+	}
+	if rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+		return fmt.Errorf("extends path escapes child template's directory: %s", parentPath)
+	}
+
+	return nil
 }
 
 // canonicalPath returns a symlink-resolved absolute path for use as a stable
--- a/internal/config/merge.go
+++ b/internal/config/merge.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


internal/config/merge.go
```diff
@@ -615,6 +616,150 @@ func validateAndFixImmutabilityConfig(template *ImageTemplate) {
...
+			if os.IsNotExist(err) {
+				return nil, fmt.Errorf("extends path not found: %s", parentPath)
+			}
+			return nil, fmt.Errorf("failed to access extends path %s: %w", parentPath, err)
+		}
+
+		if _, err := security.CheckSymlink(parentPath, security.RejectSymlinks); err != nil {
+			return nil, fmt.Errorf("invalid extends path: %w", err)
+		}
+
```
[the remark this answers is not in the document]
arodage: Addressed in dbe0cc6a. Added verifyResolvedContainment() which uses filepath.EvalSymlinks to confirm the fully symlink-resolved parent stays within the resolved child directory, so a directory symlink like child/link -> /outside can no longer bypass the lexical guard. Added TestResolveExtendsChainDirectorySymlinkEscape covering that vector.

internal/config/merge.go
```diff
@@ -615,6 +616,183 @@ func validateAndFixImmutabilityConfig(template *ImageTemplate) {
...
+			cyclePaths := append([]string{}, displayPaths[cycleStart:]...)
+			cyclePaths = append(cyclePaths, parentPath)
+			return nil, fmt.Errorf("circular extends detected: %s", formatCyclePath(cyclePaths))
+		}
+
+		currentPath = parentPath
+	}
+
+	depth := len(leafToRoot) - 1
+	if depth > 4 {
```
magerstam, line 694: We only log this as a warning, but allows for users to extend beyond 4? I think that might be fine but double checking.

internal/config/merge.go
```diff
@@ -615,6 +616,183 @@ func validateAndFixImmutabilityConfig(template *ImageTemplate) {
...
+	parentPath, err := filepath.Abs(parentCandidate)
+	if err != nil {
+		return "", fmt.Errorf("failed to resolve extends path %q: %w", extendsRef, err)
+	}
+
+	rel, err := filepath.Rel(childDir, parentPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to evaluate extends path %q: %w", extendsRef, err)
+	}
+	if rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
```
magerstam, line 730: Is this intentional, where we are forcing the templates to all exist in the same directory? Is the intention that templates must reside in the image-template directory, or that parent / child has to reside in the same directory?

magerstam (approved):
I think this looks pretty good overall. Merge logic is going into a separate PR? Only few minor comments.

internal/config/merge.go
arodage: Yes, adding this as a warning only. We will revisit if we notice any issues with the highly nested template.

internal/config/merge.go
arodage: How about following response:"It's a containment rule, not strictly same-directory: a parent must resolve within the child template's own directory subtree. Same dir (extends: base.yml) and subdirs (extends: common/base.yml) are allowed; anything above it (../base.yml), a sibling dir, or an absolute path outside that subtree is rejected. It's relative to wherever the leaf template lives, not tied to image-templates.
Intentional, for security and predictability:

Path-traversal guard: extends is user-controlled and causes us to load an arbitrary file; containment prevents reading files elsewhere on the host (e.g. ../../../secrets.yml). A post-EvalSymlinks check also blocks symlink escapes.
Portability/reproducibility: template folders stay self-contained, so the whole inheritance chain travels together and a build's inputs live in one tree.

If it's too restrictive, we can relax it

magerstam: Sure, that makes sense.

PR #724, opened 2026-07-03
Copy baseline to workspace and attach loop device

yktyagi:
First concrete overlay ingestion stage: copy the user-provided baseline RAW into the build workspace (never mutating the original), attach a loop device for downstream inspection/modification, and guarantee cleanup on every exit path. Gated on overlay mode; downstream stages (mount, OS detection, install) consume the Context populated here.

Loop primitive (internal/image/imagedisc/loopdev.go):
  - AttachImageToLoopDev attaches an already-existing image via losetup -fP and enumerates partition nodes; unlike CreateRawImageLoopDev it neither creates nor partitions the backing file, so it is safe on a baseline that must not be modified
  - loopDevPartitions enumerates partition nodes via lsblk, excluding the base device
  - on partition-enumeration failure the loop device is detached before returning, so no loop device leaks
  - AttachImageToLoopDev added to LoopDevInterface

Overlay ingestion (internal/image/overlay/overlay.go):
  - Ingestor.WithBaseline acquires (copy + loop attach), runs the downstream func, then cleans up
  - local source.path is copied via an in-process, sparse-aware streaming copy (copyLocalFile — no shell, produces a real file, never a symlink or move); source.url is downloaded over TLS via network.DownloadFile; the user-provided source is never modified. The destination is hardened before writing (copyBaseline/prepareDestination): any pre-existing entry at the fixed workspace path is unlinked so a planted symlink or hardlink cannot redirect or truncate an unintended target, and the local copy opens O_CREATE|O_EXCL to close the TOCTOU window
  - defer-based cleanup detaches the loop device on success, failure, and panic; the workspace copy is removed only on full success
  - retainCopy (defaults to debug mode) keeps the workspace copy for debugging
  - loop device path and partitions stored in Context for downstream stages; all shell ops route through the allowlist

Tests:
  - overlay_test.go (mocked loop device + shell): copy-not-symlink, detach on success/error/panic, copy removal on success, retention, attach-failure cleanup, missing-source; prepareDestination hardening (breaks a hardlink without truncating the victim, drops a symlink without following it, refuses a directory, no-op on missing dst)
  - overlay_integration_test.go: attaches/detaches a real loop device (dd -> losetup -fP); skips when not root or losetup unavailable
  - loopdev_test.go: AttachImageToLoopDev and loopDevPartitions including the detach-on-enumeration-failure path
  - rawmaker_test.go: mockLoopDev satisfies the extended interface

#### Merge Checklist  

- [ ] Ready to merge


internal/config/schema/os-image-template.schema.json
arodage, line 514: Schema defines both baseline.source.path and baseline.source.url, but it does not enforce exactly one of them. Please add oneOf under BaselineSource so templates cannot provide both or neither.

"oneOf": [
  { "required": ["path"] },
  { "required": ["url"] }
],

internal/image/overlay/overlay.go
```diff
@@ -0,0 +1,522 @@
...
+			// device is never silently swallowed. When fn already failed, join
+			// the two so both are reported; when fn succeeded, this alone marks
+			// the run unsuccessful.
+			err = errors.Join(fnErr, detachErr)
+		}
+		if !panicked && err == nil {
+			// Honor debug retention only on a fully successful build.
+			ing.removeCopy(ctx, false)
+		} else {
+			log.Debugf("Retaining workspace baseline copy after unsuccessful build: %s", ctx.BaselineCopyPath)
```
arodage, line 241: On unsuccessful runs we always retain the workspace baseline copy, regardless of retainCopy. Since baseline images are large, repeated failures can accumulate significant disk usage. Should we conditionally remove the copy when retainCopy is false, and retain only for debug mode (or panic cases)?

internal/image/overlay/overlay.go
```diff
@@ -0,0 +1,522 @@
...
+// closes the same race the local-copy path closes, so a privileged run against an
+// attacker-writable workspace cannot be redirected to clobber an unintended path.
+// The file is created 0600 up front so the baseline stays private (no post-hoc
+// chmod window), matching copyLocalFile.
+//
+// A CheckRedirect policy rejects any redirect to a non-https target: BaselineSource
+// .Validate enforces https-only on the configured URL, but Go's default client
+// follows redirects, so an https URL could be bounced to an http Location and fetch
+// the baseline over plaintext. Refusing the redirect keeps the whole fetch on TLS.
+func downloadBaseline(url, dst string) (err error) {
```
arodage, line 377: downloadBaseline streams the HTTP body directly to disk with no size validation, which creates two related problems:
- Silent truncation: if the body is cut short but the transfer reports success, we write a corrupt baseline that only fails later at mount/inspect with a confusing error.
- Unbounded download: when the server sends no Content-Length (chunked/compressed), there is no upper bound, so a misconfigured or malicious URL can stream an unbounded body and exhaust workspace disk.

Please bound and verify the download: reject early when a declared Content-Length exceeds a max cap, bound the streaming copy with io.LimitReader so an undeclared-length body cannot exceed the cap either, and verify the bytes written against Content-Length when it is >= 0. The partial copy from any error path is already cleaned up by acquire's removeCopy(ctx, true).

Add a package-level cap near the other consts at the top of file:
```
// maxBaselineBytes bounds the size of a downloaded baseline image so a
// misconfigured or malicious URL cannot exhaust workspace disk. Sized generously
// for a whole-disk RAW image; adjust if larger baselines become legitimate.
const maxBaselineBytes = 64 << 30 // 64 GiB
```

Add following check in this function:
```
    if resp.StatusCode != http.StatusOK {
        return fmt.Errorf("unexpected HTTP status %d", resp.StatusCode)
    }

    // Reject early when the server declares a length larger than the cap, so we
    // never begin writing an oversized image.
    if resp.ContentLength > maxBaselineBytes {
        return fmt.Errorf("baseline image too large: declared %d bytes exceeds limit %d", resp.ContentLength, maxBaselineBytes)
    }

    out, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0600)
    if err != nil {
        return err
    }
    defer func() {
        if cerr := out.Close(); cerr != nil && err == nil {
            err = cerr
        }
    }()

    // Bound the copy even when Content-Length is unknown (-1): read at most one
    // byte past the cap so an over-limit body is detectable rather than silently
    // truncated exactly at the cap.
    written, err := io.Copy(out, io.LimitReader(resp.Body, maxBaselineBytes+1))
    if err != nil {
        return fmt.Errorf("failed to write %s: %w", dst, err)
    }
    if written > maxBaselineBytes {
        return fmt.Errorf("baseline image too large: exceeded limit %d bytes", maxBaselineBytes)
    }

    // Guard against a silently truncated download: when the server declared a
    // Content-Length (>= 0), the bytes actually written must match it. A short
    // body otherwise yields a corrupt baseline that only fails later at
    // mount/inspect with a far less obvious error.
    if resp.ContentLength >= 0 && written != resp.ContentLength {
        return fmt.Errorf("incomplete download of %s: wrote %d of %d bytes", dst, written, resp.ContentLength)
    }
    return nil
```

arodage (approved):

internal/config/schema/os-image-template.schema.json
yktyagi: fixed

internal/image/overlay/overlay.go
yktyagi: fixed

internal/image/overlay/overlay.go
yktyagi: fixed

PR #743, opened 2026-07-13
feat(build): support --nocache option to build image from scratch

DebalGhosh:
#### Merge Checklist  

- [ ] Ready to merge

#### Description 

Adds a `--nocache` flag to the `build` command. When set, the build runs in fresh, unique cache and workspace directories instead of the configured ones, so nothing is reused from previous builds (a true from-scratch build). The unique directories are created adjacent to the configured `cache_dir`/`work_dir` (same filesystem, safe for multi-GB builds) and removed once the build finishes.

Because every subsystem resolves directories from the config singleton, overriding it once in `executeBuild` propagates to all consumers — the same mechanism the existing `--cache-dir`/`--work-dir` flags use. Since the final image is written inside the workspace, it is copied into the configured `work_dir` before the temporary directories are cleaned up, so the output is preserved. `--nocache` is mutually exclusive with `--cache-dir`/`--work-dir`; if the output copy-out fails, the unique workspace is kept for recovery.

Key changes:
- `cmd/image-composer-tool/build.go`: new `--nocache` flag; `setupNoCache()` (creates unique dirs, overrides config, returns a cleanup closure) and `preserveNoCacheOutput()` helpers; wiring in `executeBuild` with a deferred cleanup that runs after `PostProcess` unmounts the chroots (so `RemoveAll` is safe).
- Tests for the helpers, the copy-out, the error/cleanup paths, and the mutual-exclusion check.
- Docs: CLI spec flag table, usage guide, caching doc, and release notes.

None.

- `go test ./cmd/image-composer-tool/` — all tests pass, including new tests for `setupNoCache` (success, cleanup, keep-workspace-on-failure, and missing-parent error paths), `preserveNoCacheOutput` (copy + no-op when no output), and `--nocache` + `--cache-dir`/`--work-dir` mutual exclusion.
- `gofmt -l`, `go vet ./cmd/image-composer-tool/`, and `golangci-lint run ./cmd/image-composer-tool/` (v1.64.7) — all clean.
- `go build ./...` — succeeds.

### DIFF
--- a/cmd/image-composer-tool/build.go
+++ b/cmd/image-composer-tool/build.go
@@ -16,6 +16,7 @@ import (
 	"github.com/open-edge-platform/image-composer-tool/internal/provider/rcd"
 	"github.com/open-edge-platform/image-composer-tool/internal/provider/ubuntu"
 	"github.com/open-edge-platform/image-composer-tool/internal/utils/display"
+	fileutil "github.com/open-edge-platform/image-composer-tool/internal/utils/file"
 	"github.com/open-edge-platform/image-composer-tool/internal/utils/logger"
 	"github.com/open-edge-platform/image-composer-tool/internal/utils/system"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


DebalGhosh:
Superseded by #744, which is based on a branch in the main repository (`open-edge-platform:feature/nocache-build`) rather than my fork. Closing this to avoid a duplicate.

PR #760, opened 2026-07-17
fix(chroot): avoid rm -rf on shared devtmpfs chroot mountpoint

arodage:
#### Merge Checklist

- [ ] Ready to merge

#### Description

`CleanSysfs` (`internal/utils/mount/mount.go`) removed the chroot's `run`/`sys`/`proc`/`dev`
mountpoint directories with `rm -rf` whenever its mount-table lookup (a string match
against `mount` command output) believed the path was no longer mounted. It is called
from the RPM installer's chroot cleanup (`internal/chroot/rpm/installer.go`).

The chroot `dev` directory is mounted with `devtmpfs`, which is a single kernel-wide
filesystem instance shared across **every** mountpoint of that type — mounting it
fresh at a chroot path mirrors the real host `/dev`, it is not an independent copy
like `tmpfs`. If the "still mounted" check is ever wrong (e.g. a path-normalization
mismatch, or an unmount that completes lazily), the following `rm -rf` on that path
is equivalent to `rm -rf /dev` on the **host**.

**Fix:** replace `rm -rf` with `rmdir` for these mountpoint directories in `CleanSysfs`.
`rmdir` only removes an already-empty directory, so a false "not mounted" verdict now
fails loudly with an error instead of silently destroying host device nodes. `rmdir`
is added to the `internal/utils/shell` command allowlist (required — `ExecCmd` rejects
commands not in the allowlist).

This reduces the blast radius of the bug but does not change the underlying
mount-table detection logic in `GetMountPathList`/`slice.Contains`, which can still
be wrong; it just makes that wrongness fail safe instead of destructive.

**How this was found:** on a dev host, real device nodes (`/dev/null`, `/dev/ptmx`,
`/dev/zero`, `/dev/random`, `/dev/urandom`, `/dev/tty`) were repeatedly destroyed,
breaking `sudo` TTY prompts and all new terminal/pty allocation on the host. The
directly observed incidents were on an **Ubuntu (deb) build** with a stale work
directory left with `proc`/`sys`/`dev` still mounted — i.e. the same "`rm -rf` across
a shared `devtmpfs` mount" hazard, but via a manual cleanup of a stuck build
directory, not confirmed to be through this exact `CleanSysfs`/RPM code path.
Auditing the code turned up `CleanSysfs` as a real, independently verifiable instance
of this hazard, which is what this PR fixes.

#### Known follow-up (not included in this PR)

`internal/chroot/deb/installer.go`'s `cleanupOnError` unmounts only the local repo
path and then runs `rm -rf` on the whole chroot env path — it never unmounts
`proc`/`sys`/`dev` first. If those are still mounted, this has the same
shared-`devtmpfs` hazard as the bug fixed here, and is a closer match to what was
observed on the affected host. Left out of this PR to keep the change scoped and
reviewed independently; should be addressed as a follow-up.

None.

\`\`\`bash
go build ./cmd/... ./internal/...
go test ./internal/utils/mount/... ./internal/utils/shell/...
go vet ./internal/utils/mount/... ./internal/utils/shell/...
gofmt -l internal/utils/mount/mount.go internal/utils/mount/mount_test.go internal/utils/shell/shell.go
\`\`\`

All pass. `CleanSysfs` unit tests updated to assert `rmdir` (with proper shell-quoted
paths) is used instead of `rm -rf`, and to cover the still-mounted / failure paths.


internal/utils/mount/mount_test.go
```diff
@@ -870,7 +870,7 @@ func TestCleanSysfs(t *testing.T) {
 			mountPoint: "/mnt/chroot",
 			mockCommands: []shell.MockCommand{
 				{Pattern: "mount", Output: "", Error: nil},
-				{Pattern: "rm -rf", Output: "", Error: fmt.Errorf("permission denied")},
+				{Pattern: "rmdir", Output: "", Error: fmt.Errorf("permission denied")},
```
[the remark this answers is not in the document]
arodage: addressed!

internal/utils/mount/mount.go
```diff
@@ -388,7 +388,14 @@ func CleanSysfs(mountPoint string) error {
...
 		if !slice.Contains(pathList, fullPath) {
-			if _, err := shell.ExecCmd("rm -rf "+fullPath, true, shell.HostPath, nil); err != nil {
+			// Use rmdir (not rm -rf) here: "dev" is mounted with devtmpfs, which is a
+			// single kernel-wide filesystem instance shared across every mountpoint of
+			// that type. If the "still mounted" check above is ever wrong (e.g. mtab
+			// parsing mismatch), a recursive rm -rf on this path would actually delete
+			// the host's real /dev entries (e.g. /dev/null, /dev/ptmx). rmdir only
+			// succeeds on an empty directory, so a false negative fails loudly instead
+			// of destroying host device nodes.
+			if _, err := shell.ExecCmd("rmdir "+shell.QuoteArg(fullPath), true, shell.HostPath, nil); err != nil {
```
yktyagi, line 402: `rmdir` is not idempotent on a missing/non-empty dir. Not reachable via the current RPM call site, but if `CleanSysfs` is ever reused on a stale/partially-built chroot, a missing top-level dir (ENOENT) or a leftover non-mount file will now error where `rm -rf` silently succeeded. Worth a one-line note in the doc comment that `CleanSysfs` now requires the dirs to be empty (post-unmount) and will error otherwise — that's the intended contract, so make it explicit.

arodage: Addressed

yktyagi (approved):

PR #772, opened 2026-07-22
initial draft for container ADR in ICT

magerstam:
#### Merge Checklist  

- [ ] Ready to merge

#### Description 

This ADR provides a proposal to support pre-build containers, Dockerfiles and deployment of 
containers with Docker compose during the composition phase in ICT.


docs/architecture-decision-record/adr-container-support.md
```diff
@@ -0,0 +1,1123 @@
...
+
+Example:
+
+```yaml
+containers:
+  runtime:
+    type: docker
+    installPolicy: auto
+
+  images:
```
srmungar, line 221: is the url of the Dockerfile manifest based on which the container image is to be built, is listed in this schema?

docs/architecture-decision-record/adr-container-support.md
```diff
@@ -0,0 +1,1123 @@
...
+    - name: inference-service
+      source:
```
srmungar, line 223: A simple approach to build a container image is to take the inputs requierd to execute the following command docker build [OPTIONS] PATH | URL | -  ,   correct?, This section can be simplified to collect the parameters from this section to execute the docker build command later in the ICT build phase.  if so, what is the repository: filed refer to here? In order to enable the IMA team faster, we can adopt simpler approaches to build container images.

docs/architecture-decision-record/adr-container-support.md
```diff
@@ -0,0 +1,1123 @@
...
+        repository: localhost/edge/inference
+        tag: "1.0.0"
+      build:
+        target: runtime
+        args:
+          MODEL_VARIANT: small
+      lifecycle:
+        autostart: false
+
+    - name: metrics-agent
```
srmungar, line 237: consider importing a container rootfs into the chroot directly. Perhaps the approach taken here can be pushed to next phase of implementation for faster enablement of IMA.

PR #786, opened 2026-07-29
fix(webui): eliminate false cancellation-failure and automate sudoers setup

Shruti-Raj-Vansh-Singh:
## Why

The build-cancellation flow merged in #781 fails for people validating it from a fresh `main` in two distinct ways. This PR fixes both, and removes the manual sudoers setup that was the most common cause of "it doesn't work on my box".

## What breaks today

1. **A successful cancel reports a false `cancellation-failure`.** `signalCancel` treated *any* non-zero `sudo -n kill` exit as a delivery failure. But the build's group leader is the `sudo` wrapper, so signalling `-pgid` races that wrapper's exit: `TERM` reaches ICT, which tears down and exits `130`, yet `kill` then walks a group whose member has already gone and exits non-zero (`ESRCH`, often with no output). Result: the UI raises a "machine may need manual cleanup" alarm on a build that cancelled and tore down cleanly.

2. **Cancel could take down the server (and your SSH session).** `runBuild` recorded the pgid as `cmd.Process.Pid` *before* the child-side `setpgid` necessarily completed. A cancel serviced in that window could deliver `kill -TERM -<pgid>` to the **server's own process group** — killing the server, and when it was launched from an interactive SSH login, the whole session with it.

3. **Sudoers setup is manual and error-prone.** `--sudo` needs three scoped NOPASSWD rules (build, kill, cat). A wrong path or a missing rule silently breaks cancellation and downloads. People commonly grant only the `build` rule.

## Changes

**Cancellation correctness**
- `isSudoAuthFailure()` distinguishes a genuine sudo-auth failure (missing rule / password required) — surfaced as a real `cancellation-failure` — from the benign teardown race, which is swallowed. The child's actual exit (`classifyExit`) and the watchdog are the source of truth.
- `runBuild` reads the child's real group with `Getpgid` instead of assuming `pid == pgid`.
- `signalCancel` hard-refuses to signal a pgid that matches the server's own group.
- New `signalGroup` seam on `Server` so tests exercise the cancel paths without a real `sudo kill`.

**Automated sudoers setup**
- `serve --print-sudoers` emits the exact three rules for the current user, resolved binary, and workspace — resolving `kill`/`cat` the way `sudo` does at runtime and scoping the `cat` rule to the workspace subtree (not any root-owned path).
- `scripts/install-sudoers.sh` generates + `visudo`-validates + installs atomically to `/etc/sudoers.d`, as the target user.

**Docs**
- `serve.go --sudo` help and `web/README.md` updated to three rules, leading with the automated path, and recommending `serve` run in its own session (`setsid`/systemd) so a stray group signal can never reach the operator's shell.

## Tests

`go test ./internal/api/... ./cmd/image-composer-tool/...` passes. New:
- `TestIsSudoAuthFailure`, `TestSignalCancelRefusesOwnGroup`
- `TestRunBuildCancelClassifiesCancelledNoResidual` — in-process end-to-end: a real cancelled child is classified `cancelled` with **no residual**
- `TestSudoersSpecRender`, `TestResolveSudoersSpecResolvesRelativePaths`

Generated rules verified against `visudo -cf`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Shruti-Raj-Vansh-Singh:
Thanks for the review. I have addressed all of the Copilot comments:

1. resolveSudoHelper did not match sudo's secure_path (sudoers.go)

It now resolves the kill and cat helpers against sudo's own secure_path, which it reads via `sudo -n -l`, and falls back to sudo's compiled-in default (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin) when that cannot be read. I removed the exec.LookPath ($PATH) fallback entirely, because matching the caller's $PATH is exactly the runtime mismatch and unsafe-writable-directory risk this rule needs to avoid. While testing on a fresh machine I found that `sudo -l` escapes the ':' separators as '\:', so I added parseSecurePath (and TestParseSecurePath) to handle that correctly.

2. Non-auth `sudo -n kill` failures were being swallowed (builds.go)

I inverted the classification. Previously any non-auth failure was treated as a harmless race. Now only the genuinely-already-gone case ("no such process", ESRCH) is treated as delivered, and every other non-zero exit (command not found, unable to execute, empty output, misusage) is reported as a real cancellation failure. I renamed isSudoAuthFailure to isKillTargetGone, and TestIsKillTargetGone now asserts those cases are treated as real failures.

3. The cat rule was scoped too broadly (sudoers.go, help text, README)

I scoped it to <work-dir>/builds/* instead of the whole work dir. I confirmed all artifacts live under <work-dir>/builds/<id>/... (see handleStartBuild and history.go). I renamed the SudoersSpec.WorkDir field to BuildsDir and updated its comment, and I updated the test fixtures, the --sudo help text, and the README example and prose to match.

I also added tests for the new `serve --print-sudoers` path and the signalCancel branches. I validated everything end to end from a fresh clone: build, the full internal/api and cmd test suites, and the generated rules pass `visudo -cf`.

PR #796, opened 2026-08-03
refactor(templates): categorize image templates by distribution and compose them with extends

DebalGhosh:
#### Merge Checklist  

- [x] Ready to merge

#### Description 

`image-templates/` held 60 templates in one flat listing. This groups them by
distribution and then uses `extends:` to express the derived images as deltas
instead of copies.

**Independent of #790.** That PR is still open and adds its own composition
examples; this one touches none of its files and uses a different filename for
its robotics example. When #790 merges, its four templates land flat in
`image-templates/` and need a one-line `git mv` into `ubuntu24/` — all three
levels of its chain stay siblings, so the chain keeps resolving.

##### 1. Grouped by distribution

```
image-templates/{azl3,debian13,el10,elxr12,elxr13,emt3,ubuntu24,ubuntu26}/
```

Filenames are unchanged. Directories are named after `target.dist`, not the
filename prefix, so `rcd10-*.yml` (which declare `dist: el10`) are in `el10/`
and `elxr-edge-26.04-*.yml` (which declare `dist: elxr13`) are in `elxr13/`.

Distribution is the axis because it is the only grouping that cannot break
inheritance. `resolveExtendsParentPath` resolves a parent relative to the
child's own directory and rejects anything that escapes it, so every level of a
chain must be a sibling; `targetsMatch` additionally requires all levels to
share `os`/`dist`/`arch`/`imageType`. A grouping drawn from those four fields
can therefore never separate two templates that were allowed to form a chain.
Grouping by purpose would — a chain legitimately runs minimal → robotics.

The first commit is a pure move: no template content changes, only the 24
`scripts/build_*.sh` paths, two tests, `internal/api/data/manifest.yaml` and
docs. Done with `git mv`, so all 58 files are rename-detected and in-flight PRs
rebase cleanly.

##### 2. Derived templates now inherit instead of duplicating

```
emt3-x86_64-edge-raw.yml              41 packages
  → emt3-x86_64-emf-raw.yml           + persistent /opt, verity root, 2 pkgs
  |   → emt3-x86_64-emf-rt-raw.yml    + real-time GPU driver
  → emt3-x86_64-dlstreamer.yml        + media/AI stack, 3 repos, 8GiB disk

debian13-x86_64-minimal-raw.yml
  → debian13-x86_64-bb-raw.yml        + initramfs-tools, 3 hook files, 3 cmds

ubuntu24-x86_64-minimal-raw.yml
  → ubuntu24-x86_64-robotics-core-raw.yml            ROS 2 Jazzy, OpenVINO, Gazebo
      → ubuntu24-x86_64-robotics-slam-raw.yml        + collab-SLAM, RealSense, NPU
```

And a second robotics route that does **not** build an OS from scratch — overlay
and extends composed together:

```
Canonical noble cloud image (qcow2, never modified in place)
  → ubuntu24-x86_64-robotics-hw-overlay-qcow2.yml        Intel oneAPI / Level Zero / NPU / RealSense
      → ubuntu24-x86_64-robotics-jazzy-overlay-extends.yml  + ROS 2 Jazzy, OpenVINO, Gazebo, SLAM
```

This removes a 41-package block that was copied verbatim into four EMT3
templates, comments and blank lines included. The three EMT3 templates go from
157/160/226 lines to 103/93/120; `debian13-x86_64-bb-raw.yml` goes from 135 to 64.

`debian13-x86_64-bb-dracut-raw.yml` deliberately stays standalone and now says
why: the minimal image declares `cloud-initramfs-growroot`, which
`Depends: initramfs-tools`, and packages cannot be removed by a child.

##### 3. Two defects found while doing this

**Repositories sharing a `codename` collapse across layers.**
`mergePackageRepositories` overrides the first codename match instead of
appending, so duplicate codenames survive only inside one layer. The robotics
templates declare three repositories all codenamed `noble` (ROS 2, Gazebo,
RealSense). Extending a parent that declares any repository of its own collapses
them to one and `resolve --full` shows five repositories where there should be
seven — **ROS 2 and Gazebo silently gone**. Minimal reproduction: a parent with
one `noble` plus a child with two resolves to exactly one.

This is why the robotics example extends the *minimal* image, which declares no
repositories (as do the OS defaults), and why the whole repository block is kept
in one layer. It is worth fixing in `internal/config/merge.go` independently;
#790 reached the same conclusion.

**Template discovery was not recursive.** `ScanTemplates` skipped directory
entries, so grouping templates into subdirectories would have silently emptied
the AI/RAG index and the web UI template list with no error. Fixed here, along
with the three Copilot skill scripts (`glob` → `rglob`). Verified: 63 templates
discovered after, 0 before.

##### 4. Pre-existing bugs fixed in passing

- `elxr-cloud-amd64.yml` referenced `files/etc/...` while the files ship in
  `elxr-cloud-amd64/files/etc/...`; all four `additionalFiles` were silently
  dropped on `main`.
- `ubuntu24-x86_64-robotics-jazzy-raw.yml` pinned `librealsense2` where its ISO
  sibling pinned `librealsense2*`, leaving the sub-packages unpinned.
- Three documentation links named template files that do not exist.

##### 4b. Kernel-version fix folded in (was PR #802)

PR #802 is superseded by this branch and closed. Four ubuntu24 OS defaults under
`config/osv/ubuntu/ubuntu24/imageconfigs/defaultconfigs/` pinned
`kernel.version: "6.17"` next to the rolling metapackage
`linux-image-generic-hwe-24.04`, and noble no longer ships 6.17 — so
`ubuntu24-x86_64-minimal-initrd.yml` and `-dkms-demo.yml` failed in
pre-processing even though the templates themselves were clean, because a
template with no `kernel.version` of its own inherits the default's.

#765 removed that antipattern from the 12 affected user templates but never
touched the OS defaults, which is why the failure recurred after #494, #669 and
#761. `robotics-demo-ubuntu24-x86_64.yml` was also missed there and is fixed
here. Verified by building: `minimal-initrd` (3m22s) and the default config
directly (2m58s) both complete with zero `kernel version mismatch` errors.

##### 4c. Robotics composed from the Canonical cloud image (overlay + extends)

Two new templates show the composition features used *together* rather than
building an OS from scratch — the pattern from commit `24192fae`, applied to
Ubuntu robotics:

- `ubuntu24/ubuntu24-x86_64-robotics-hw-overlay-qcow2.yml` — overlay base:
  layers Intel hardware enablement (oneAPI runtime, Level Zero, NPU drivers,
  RealSense DKMS) onto Canonical's official noble cloud image, emits a 24GiB
  qcow2, and never modifies the vendor image.
- `ubuntu24/ubuntu24-x86_64-robotics-jazzy-overlay-extends.yml` — `extends` that
  base to add ROS 2 Jazzy, OpenVINO nodes, Gazebo Harmonic and collaborative SLAM.

The child declares **no** `baseline`, `overlayPolicy`, `packageRepositories` or
`disk`. That is load-bearing, not stylistic: `MergeConfigurations` replaces the
package list outright when the merged template is itself overlay-mode, and
`IsOverlayMode()` is true exactly when `baseline` is non-nil. Verified both ways —
a child without `baseline` yields `[parent…, child…]`; a child that redeclares it
yields `[child…]` and the parent's packages vanish silently.

Two further consequences the split relies on: parent packages are *prepended*, so
the order-coupled `intel-oneapi-runtime-*` trio in the base resolves ahead of the
child's `ros-jazzy-desktop` by construction; and all seven repositories stay in the
base because three share the codename `noble` and would collapse if split.

`resolve --full` on the child confirms: baseline/policy/disk inherited, 7
repositories with all three `noble` entries intact, 19 packages with oneAPI at
indices 0-2 ahead of `ros-jazzy-desktop` at 11, no create-mode package leakage, 10
configurations ordered base-first. Overlay preflight against the real Canonical
image reports **59 add, 0 upgrade, 0 downgrade, 0 conflict, 0 unsatisfiable** — the
additive-only fail-closed gate accepts the stack cleanly.

> **Both tool defects this branch surfaced now have resolution PRs off `main`:**
> **#804** (`lsblk` START fallback to `sfdisk`, unblocks resize on Ubuntu 22.04) and
> **#805** (batch `dpkg -i` under `MAX_ARG_STRLEN`, unblocks the ROS 2 child).
> Neither touches a template, so nothing here needs rebasing on them.

**Not taken to a finished artifact on this host, and why.** The resize path runs
`lsblk -o PATH,START,TYPE`; this host is Ubuntu 22.04 (util-linux 2.37.2) and the
`START` column landed in util-linux 2.38, so it fails with
`lsblk: unknown column: START,TYPE`. The repo recommends Ubuntu 24.04, and the base
template's header records the requirement. Worth noting separately:
`internal/image/overlay/resize.go:239` checks that `lsblk` *exists* but not that it
supports the column it needs — a capability check (or an `sfdisk -d` fallback)
would make overlay resize portable to 22.04. Filed as a follow-up rather than
changed here, since it is a Go change to a safety-critical guard.

The guard itself would pass for this baseline: read from the published image, root
starts at sector 2099200 while the ESP starts at 10240, so root is the last
partition and growing it cannot run into a following one.

##### 5. Documentation

Three new guides ship in `image-templates/` — follow the links to read them rendered:

| Guide | What it covers |
|---|---|
| **[image-templates/README.md](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/README.md)** | The catalog: every template by distribution, arch, image type and purpose; how to choose a starting point; and which templates CI actually builds |
| **[image-templates/COMPOSITION.md](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/COMPOSITION.md)** | `extends:` and overlay mode — the two rules they enforce, the per-section merge table, and the four traps that cost real debugging time |
| **[image-templates/CONVENTIONS.md](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/CONVENTIONS.md)** | Naming, required sections, mandated key order, and how to add a template or a new distribution |

Plus a `README.md` per distribution, each with that family's inheritance edges and
CI coverage: [ubuntu24](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/ubuntu24/README.md) · [emt3](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/emt3/README.md) ·
[debian13](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/debian13/README.md) · [elxr12](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/elxr12/README.md) ·
[elxr13](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/elxr13/README.md) · [azl3](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/azl3/README.md) ·
[el10](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/el10/README.md) · [ubuntu26](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/ubuntu26/README.md)

The per-distribution tables are generated from the templates themselves rather than
hand-written, so counts, targets, inheritance edges and CI coverage match the tree.
`.github/instructions/image-templates.instructions.md`, the root README and the
release notes are updated too.

[COMPOSITION.md](https://github.com/open-edge-platform/image-composer-tool/blob/9058df4b/image-templates/COMPOSITION.md) is the one worth reading if you review nothing
else: it is what makes these features usable without reading
`internal/config/merge.go`. It also records what templates alone cannot fix — `-raw`
and `-iso` siblings cannot share a parent because `targetsMatch` compares
`imageType`, which is why the robotics raw/iso duplication is still present.

None.

**Real image build of a refactored template.**
`debian13/debian13-x86_64-bb-raw.yml`, which now inherits the minimal image,
built successfully in 4m39s → `bb-os-image-debian-13.0.raw.gz` (404 MB), `.vhdx`
and SBOM. Mounted the result and confirmed the inheritance actually took effect:

```
PRESENT /usr/local/sbin/hello.sh (755)
PRESENT /etc/initramfs-tools/hooks/hello (755)
PRESENT /etc/initramfs-tools/scripts/init-bottom/hello (755)
ii  initramfs-tools 0.148.4   (child's package installed)
PRESENT /etc/apt/preferences.d/no-systemd-boot   (child's config command applied)
```

**Equivalence of every refactor**, by diffing `resolve --full` before and after.
For all four derived templates the package list, `disk`, `image`, `target`,
`bootloader`, `immutability`, `kernel` and `configurations` compare equal. The
differences and why each is benign:

- The three EMT3 children inherit the parent's three placeholder repository
  entries (`company-internal`, `dev-tools`, `intel-openvino`), so `emf-raw` and
  `emf-rt-raw` resolve to 3 repositories where they declared none, and
  `dlstreamer` to 6 rather than 3. Inert on two counts: their URL is the literal
  `<URL>`, which `rpmutils/download.go:173` skips before any fetch, and EMT3 is
  RPM-based so apt-source generation never runs for it. Raised by @yktyagi in
  review; each template now records this in its header, and the release note and
  `COMPOSITION.md` no longer claim the merged output is unchanged without
  qualification.
- `initramfs-tools` moves from index 41 to 68 in the declared list, because
  child packages append after the parent's. Install order is recomputed by
  `pkgsorter`'s dependency-driven topological sort, so declaration order does
  not decide what gets installed.
- `additionalFiles` order changes, which `merge.go` produces from a map and the
  template documentation already records as unstable. Same destinations.
- The robotics example installs six GRUB packages from its parent that it does
  not use. Documented in the template; unavoidable without a removal syntax.

The `ORDER-COUPLED` `intel-oneapi-runtime-*` entries still resolve ahead of
`ros-jazzy-desktop`, which the version pins depend on.

**Validation**: all 62 templates schema-validate. The one failure,
`ubuntu24-x86_64-fde-raw.yml`, fails identically on `main` — its
`fde.passphraseFile` points at `/home/user/test-fde.txt`, an absolute path that
does not exist.

**additionalFiles resolution**, by comparing declared against resolved entry
counts across the whole tree: `main` drops entries in 8 templates, this branch
in 7 — the same set minus `elxr-cloud-amd64`. No regressions, one fix. The
remaining 7 are pre-existing and untouched.

**Path integrity**: every `image-templates/...yml` path in `scripts/*.sh`
resolves to a real file.

**Gates**: `go build ./cmd/... ./internal/...`, `go vet`,
`go test ./cmd/... ./internal/...` all pass; `gofmt` clean.

> Note: only 24 of the 61 templates are built by CI, so the other 37 are covered
> here by `validate` and `resolve` rather than by a build.


DebalGhosh:
### CI note: `build-ubuntu24-immutable` failure is pre-existing

This job fails on `main` as well, including at `32139ba0` — the exact commit this branch is based on — and it also failed at `8910f463`. It is not caused by this PR.

Cause, from the job log:

```
WARN  debutils/download.go:965  requested package '"linux-image-generic"' not found in repo
```

That package is requested by `ubuntu24-x86_64-edge-raw.yml`, which this PR only relocates. `git diff -M` records it as `R100` with a `0 0` numstat, i.e. a pure rename with no content change:

```
0  0  image-templates/{ => ubuntu24}/ubuntu24-x86_64-edge-raw.yml
```

Every other build job passes — 20 of 21 — including all four `build-emt3-*` jobs, which build the templates this PR converted to `extends` chains, and `build-ubuntu24-raw`, which exercises the parent of the new robotics example.

DebalGhosh:
### Ubuntu robotics chain now nests three levels, matching #790

The first revision of this PR gave the EMT3 family a three-level chain but left Ubuntu robotics at two, so the Ubuntu leaf was still a 253-line near-copy. Fixed:

```
ubuntu24-x86_64-minimal-raw.yml
  -> ubuntu24-x86_64-robotics-core-raw.yml      ROS 2 Jazzy, OpenVINO, Gazebo, 7 repos
       -> ubuntu24-x86_64-robotics-slam-raw.yml  + collab-SLAM, RealSense, NPU
```

`robotics-core` is a complete buildable image (37 packages) *and* a reusable parent. The leaf declares only `extends`, `image`, `metadata`, `target` and a `systemConfig` with name, description, 7 packages and 2 commands — **42 non-comment lines against 151** for the standalone `robotics-jazzy-raw.yml`, which is retained for comparison.

Verified with `resolve --full` against the pre-split template: 44 packages (same set), all **7** repositories with the same URLs in the same order, 12 configurations, and `bootloader`/`immutability`/`kernel`/`network`/`users` all equal. Only `image.name` and `disk.name` differ — the leaf inherits `disk` wholesale, and `disk.name` is only logged (`merge.go:946`), with build paths keying on `systemConfig.name`. The `ORDER-COUPLED` `intel-oneapi-runtime-*` entries still resolve ahead of `ros-jazzy-desktop`.

### Correction to this PR's own documentation

The earlier `COMPOSITION.md` said that splitting `packageRepositories` across layers collapses same-codename entries. That was stated without its precondition. `mergePackageRepositories` short-circuits:

```go
// internal/config/merge.go
if len(userRepos) == 0 { return defaultRepos }   // parent list passes through INTACT
if len(defaultRepos) == 0 { return userRepos }
```

So the collapse requires **two** layers to declare repositories. Duplicate codenames are safe when only one layer declares them — which is why #790's three-level chain keeps all seven repositories, and why this chain declares the whole set in `robotics-core` and none in the leaf. The RealSense repository therefore lives in `robotics-core` even though only the SLAM layer installs RealSense packages. `COMPOSITION.md` now states the precondition and documents the safe pattern.

### Relationship to #790

Still independent — different filenames, none of #790's files touched. Both PRs now add a Ubuntu robotics core/leaf pair, so whichever merges second will have a redundant pair to drop. Happy to rename these to #790's filenames if maintainers prefer this PR to supersede that part of it.

.github/skills/image-composer-build/SKILL.md
```diff
@@ -229,33 +229,33 @@ gunzip -t <artifact>.gz
...
 Output: `*.raw.gz` + `*.vhdx`
 
 ### Debian 13 minimal raw image
 
 ```bash
-sudo -E ./image-composer-tool build image-templates/debian13-x86_64-minimal-raw.yml
+sudo -E ./image-composer-tool build image-templates/debian13/debian13-x86_64-minimal-raw.yml
 ```
 
 ### EMT3 Desktop Virtualization ISO
```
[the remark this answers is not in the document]
DebalGhosh: Fixed in [f842ed8c](https://github.com/open-edge-platform/image-composer-tool/commit/f842ed8c).

You are right that the heading and the command disagreed. The command is correct and the heading was wrong: EMT3 has no desktop-virtualization template — `image-templates/emt3/` contains only `dlstreamer`, `edge-raw`, `emf-raw`, `emf-rt-raw`, `minimal-initrd`, `minimal-iso` and `minimal-raw`. The template being built is the Debian 13 one, so the heading is now "Debian 13 Desktop Virtualization ISO".

Two related inaccuracies in the same file, both pre-existing, fixed while here:

- The EMT3 row of the distro table listed `desktop-virtualization-iso` and `elvin-emf`. Neither exists under `image-templates/emt3/`, so both are removed.
- The line explaining how a template name maps to a file still said `image-templates/`. It now names the per-distribution layout, including the two cases where the directory does not match the filename prefix (`rcd10-*.yml` in `el10/`, `elxr-edge-26.04-*.yml` in `elxr13/`).

internal/config/validate/validate_test.go
```diff
@@ -25,7 +25,7 @@ func loadFile(t *testing.T, relPath string) []byte {
 
 // Test new YAML image template format
 func TestValidImageTemplate(t *testing.T) {
-	v := loadFile(t, "../../image-templates/azl3-x86_64-edge-raw.yml")
+	v := loadFile(t, "../../image-templates/azl3/azl3-x86_64-edge-raw.yml")
```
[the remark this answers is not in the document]
DebalGhosh: Not changed, because the premise does not hold for this file — but thank you, it was worth checking.

`internal/config/validate/validate_test.go` has **40** top-level `func Test...` functions and **none** of them call `t.Parallel()`. There are exactly three `t.Parallel()` calls in the file, all inside table-driven subtest closures:

```
$ awk '/^func Test/{fn=$2} /t\.Parallel\(\)/{print NR": "fn}' internal/config/validate/validate_test.go
420: TestInvalidNonWSL2TemplateWithTarArtifact(t
1246: TestValidateImageTemplateJSON_AutoExpand(t
1318: TestValidateUserTemplateJSON_AutoExpand(t
```

Each is the `t.Run(tt.name, func(t *testing.T) { t.Parallel(); ... })` idiom, i.e. it parallelizes the *subtests* of one function, not the function itself. So adding `t.Parallel()` to `TestValidImageTemplate` would make it the only top-level parallel test in the file — the opposite of consistent.

It is also outside this PR's scope: the change here is two lines, updating a fixture path for the new directory layout.

```diff
-	v := loadFile(t, "../../image-templates/azl3-x86_64-edge-raw.yml")
+	v := loadFile(t, "../../image-templates/azl3/azl3-x86_64-edge-raw.yml")
```

Making all 40 tests in this file parallel would be a reasonable cleanup, but it belongs in its own change where the shared-state implications can be reviewed properly.

DebalGhosh:
## Review feedback addressed, plus a real build and QEMU boot

### Copilot's three findings

**1. Outdated template path in a template header** — fixed in [f842ed8c](https://github.com/open-edge-platform/image-composer-tool/commit/f842ed8c). This one had no inline anchor, so I swept for it: three template headers still carried the pre-move flat build command, meaning a reader copying that line would get a "no such file" error.

```
image-templates/ubuntu24/ubuntu24-x86_64-robotics-jazzy-raw.yml:10
image-templates/ubuntu24/ubuntu24-x86_64-robotics-jazzy-iso.yml:10
image-templates/ubuntu24/robotics-demo-ubuntu24-x86_64.yml:8
```

A repo-wide sweep for `image-templates/<file>.yml` references that do not resolve now returns only two, both deliberate: the "save to" placeholder in `ai-template-generation.md`, and the release-notes line that deliberately shows the old path beside the new one.

**2. Mismatched skill-doc heading** — fixed in [f842ed8c](https://github.com/open-edge-platform/image-composer-tool/commit/f842ed8c); see [the inline reply](https://github.com/open-edge-platform/image-composer-tool/pull/796#discussion_r3710193709). The heading was wrong, not the command, and two adjacent pre-existing inaccuracies in the same table are corrected too.

**3. Missing `t.Parallel()`** — not changed, with reasoning in [the inline reply](https://github.com/open-edge-platform/image-composer-tool/pull/796#discussion_r3710195080). None of the 40 top-level tests in that file call `t.Parallel()`; the three calls are all inside table-driven subtest closures. Adding it to this one function would make it the only parallel top-level test in the file.

### Real build and QEMU boot

`ubuntu24-x86_64-minimal-raw.yml` — the root of the refactored Ubuntu robotics chain, and the parent whose six GRUB packages the chain inherits — was built and booted end to end:

```
image build completed successfully        Total Time  3m53s
  minimal-os-image-ubuntu-24.04.raw.gz   (1.05 GB)
  minimal-os-image-ubuntu-24.04.vhdx     (1.70 GB)
  spdx_manifest_deb_...json
```

Booted under UEFI (OVMF, q35, TCG — KVM is not available in this environment):

```
Linux version 6.17.0-41-generic ... #41~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC
systemd[1]: systemd 255.4-1ubuntu8.16 running in system mode
systemd[1]: Detected virtualization qemu.
debalgho-desk login:
```

7 targets reached, **0** `Failed to start` units, 0 kernel panics, reached a login prompt. The only OVMF "failed to load" line is it probing a DVD-ROM that does not exist, which is expected with no CD attached.

Earlier in this PR I also built `debian13-x86_64-bb-raw.yml`, the other template converted to `extends`, and mounted the result to confirm inheritance actually reaches the image: all three inherited `additionalFiles` present at mode 755, the child's `initramfs-tools` installed, and the child's apt-pin applied.

### What could not be booted, and why

The two robotics templates cannot complete a build in this environment. `resolve --full` proves them equivalent to the standalone template, but the build stops in package download:

```
  missing: ["linux-image-generic-hwe-24.04"]
```

This is the same pre-existing failure class as the red `build-ubuntu24-immutable` check, which fails on `main` at this PR's base commit `32139ba0` (there the missing package is `linux-image-generic`). It is not introduced here:

- `linux-image-generic-hwe-24.04` is declared identically on `main` in `ubuntu24-x86_64-robotics-jazzy-raw.yml`, and in **10** other ubuntu24 templates.
- A `Missing_Requested_Packages` report from 2026-07-21, well before this branch, records the same failure for `linux-image-generic`.
- Both packages *do* exist in `noble/main`, and the provider config already includes that component, so this is an archive/metadata resolution problem in this environment rather than a template defect.

The robotics templates are not built by CI either, which is why the PR body flags that 37 of 62 templates are covered by `validate` and `resolve` rather than by a build.

docs/user-guide/release-notes.md
```diff
@@ -92,8 +92,18 @@
...
...6_64-bb-raw.yml` extends `debian13-x86_64-minimal-raw.yml`. This removes a 41-package block that had been copied verbatim into four EMT3 templates. The Ubuntu robotics stack is expressed as a three-level chain: `ubuntu24-x86_64-robotics-core-raw.yml` extends `ubuntu24-x86_64-minimal-raw.yml` and provides ROS 2 Jazzy, OpenVINO and Gazebo as a buildable image that is also a reusable parent, and `ubuntu24-x86_64-robotics-slam-raw.yml` extends that to add collaborative SLAM, RealSense and Intel NPU support in 42 non-comment lines. The standalone `ubuntu24-x86_64-robotics-jazzy-raw.yml` is retained for comparison. Note that all seven package repositories are declared in the core layer: repositories merge by `codename` and three of them share the codename `noble`, so they collapse if two layers in a chain both declare repositories — a descendant that declares none inherits the full list intact. Each derived template was verified with `resolve --full` to produce the same functional fields as before. Note that because package lists are a union with no removal syntax, a derived template also installs its parent's packages — the robotics example documents the six GRUB packages this costs it.
```
yktyagi: This bullet states "Each derived template was verified with `resolve --full` to produce the same functional fields as before." That holds for the robotics and Debian chains, but not quite for the EMT3 family.

`emt3-x86_64-emf-raw.yml`, `emt3-x86_64-emf-rt-raw.yml` and `emt3-x86_64-dlstreamer.yml` now inherit three `packageRepositories` from `emt3-x86_64-edge-raw.yml` (`company-internal`, `dev-tools`, `intel-openvino`, all with placeholder `<URL>`) that the standalone versions did not declare, so their `resolve --full` output genuinely differs from before.

It's inert at build time — placeholder GPG keys are skipped, EMT3 is RPM-based so apt-source generation doesn't run, and `edge-raw` already ships these repos in CI — so the built image is unaffected. But since the robotics chain's extra GRUB packages get an explicit callout here, consider either softening the blanket "same functional fields" claim or noting the inherited `edge-raw` repositories for the EMT3 templates too.

DebalGhosh: You're right, and thank you — this was a real inaccuracy in the release note, not just a wording nit. Fixed in [fd10fc8a](https://github.com/open-edge-platform/image-composer-tool/commit/fd10fc8a).

I resolved the pre- and post-refactor templates to get the exact numbers:

| Template | before (`173d0720`) | after |
|---|---|---|
| `emt3-x86_64-emf-raw.yml` | **0** repositories | 3 |
| `emt3-x86_64-emf-rt-raw.yml` | **0** repositories | 3 |
| `emt3-x86_64-dlstreamer.yml` | 3 (`EdgeAI`, `OpenVINO`, `mariner`) | **6** |

So the blanket "same functional fields" claim was wrong for exactly the three templates you named.

Your inertness reasoning checks out too, and I verified both halves independently since the wording of the fix depends on them:

- `internal/ospackage/rpmutils/download.go:173` — `if repo.URL == "" || repo.URL == "<URL>" { continue }`, so placeholder repositories are skipped before any fetch.
- `GenerateAptSourcesFromRepositories` has **no** call site under `internal/provider/emt/`. EMT3 being RPM-based isn't just incidental — apt-source generation never runs for it at all.

You offered "either soften the claim or note the inherited repositories", and I did both, because the sharper point in your comment was the *inconsistency*: the robotics template documents its inherited-GRUB-packages cost in its own header while these three documented nothing. Same class of cost, disclosed in one place and not the other.

1. **`docs/user-guide/release-notes.md`** — now names both exceptions (the robotics GRUB packages and these repositories) instead of claiming the merged output is unchanged.
2. **The three EMT3 templates** — each header now records what it inherits and why it is inert, matching the pattern `ubuntu24-x86_64-robotics-core-raw.yml` already follows.
3. **[`COMPOSITION.md`](https://github.com/open-edge-platform/image-composer-tool/blob/fd10fc8a/image-templates/COMPOSITION.md)** — your comment sent me looking, and the same over-broad claim was in the "splitting a template safely" checklist: *"Expect differences only in `metadata`, `image.name` and `disk.name`."* That is the guidance a contributor actually follows when splitting a template, so it was the more consequential of the two. It now lists inherited `packageRepositories` as an expected fourth difference, with the caveat that the diff is benign **only** when the added entries are placeholders or already present — if they are real repositories, the resolver's view changed and so did the image. Check the URLs, not just the count.

The commit touches comments and prose only; no template field changes, and the resolved counts are still 3, 3 and 6.

One thing I deliberately left out of scope: deleting those three placeholder entries from `emt3-x86_64-edge-raw.yml` would make the original claim literally true and remove dead configuration. But `edge-raw` is CI-built and otherwise untouched by this PR, so changing it belongs in its own change. Happy to open a follow-up if you think it is worth doing.

Also rebased onto `79f2653b` — worth noting for anyone re-reviewing that the API service-layer split moved `internal/api/data/manifest.yaml` to `internal/api/service/data/manifest.yaml`; git carried this PR's nested-path edit across the rename, and `safeTemplatePath` (now `internal/api/service/manifest.go:109`) still accepts relative subpaths, so the manifest values remain valid.

DebalGhosh:
### Build + boot results for the overlay robotics chain

Ran on an Ubuntu 24.04.3 host (util-linux 2.39.3, KVM available), since the resize
path needs util-linux >= 2.38 — see the note at the end.

#### Overlay base: builds and boots

```
ubuntu24-robotics-hw-overlay-base-24.04.qcow2    1.67 GB, 4m11s
  preflight : 59 add, 0 upgrade, 0 downgrade, 0 conflict, 0 unsatisfiable
  resize    : 3.5 GB -> 24 GiB  (25769803776 bytes, confirmed in the artifact)
```

Booted under QEMU/OVMF with KVM:

```
Linux version 6.8.0-136-generic          <- the cloud image's own kernel; this template pins none
systemd[1]: systemd 255.4-1ubuntu8.16 running in system mode
systemd[1]: Detected virtualization kvm.
Welcome to Ubuntu 24.04.4 LTS!
... login:
```

**29 targets reached, 0 `Failed to start` units, 0 kernel panics.** Mounted the
artifact afterwards to confirm the overlay actually landed rather than trusting the
build log: all eight overlaid packages installed (`libze1`, `libze-intel-gpu1`,
`intel-level-zero-npu`, `intel-driver-compiler-npu`, `librealsense2-dkms`,
`intel-oneapi-runtime-opencl`, `linux-headers-generic`, `linux-firmware`), the three
apt pins written (`eci-flann-block`, `oneapi`, `librealsense`),
`10-intel-vpu.rules` present, and the SBOM at `/usr/share/sbom/`.

#### ROS 2 child: one real fix, one tool bug

**Fixed in [b185552f](https://github.com/open-edge-platform/image-composer-tool/commit/b185552f).** Preflight blocked with exactly one violation:

```
unsatisfied-dependency "libudev-dev": requires
libudev1 (= 255.4-1ubuntu8.16-ecir8+etf+taprio+mqprio+fpe+ffw) but the post-install
set has 255.4-1ubuntu8.16 and no satisfying version is being installed
[rule: unsatisfiable-versioned-dependency]
```

ECI ships systemd with real-time networking patches, and its `libudev-dev` needs the
matching patched `libudev1`; installing that replaces the cloud image's stock copy,
which is an upgrade, so `additive-only` refused it. Preferring ECI's systemd is the
reason that repo carries priority 1000 — the ETF/taprio/mqprio/FPE queueing
disciplines are what make the stack usable for time-sensitive traffic, and the
from-scratch template gets them implicitly by building the OS itself. The base now
declares `packageOperation: additive-and-upgrade`; downgrades and removals stay
blocked and `conflictPolicy: fail` is unchanged. Preflight then passes: **2002 add,
2 upgrade, 0 downgrade, 0 conflict, 0 unsatisfiable, 0 policy violations**.

**Not fixed here — a scalability bug in the overlay installer.** With the gate
cleared, all 2004 packages resolve, download and verify, then install fails before
`dpkg` starts:

```
dpkg install of 2004 artifact(s) failed: failed to start command …
argument list too long
```

`internal/image/overlay/install.go` passes every artifact path in a single `dpkg -i`
invocation. At this scale the command string reaches **141932 bytes**, over the
**131072-byte** per-argument ceiling (`MAX_ARG_STRLEN` = `PAGE_SIZE × 32`) that
applies because commands run through `bash -c`. `ARG_MAX` (2 MiB) is *not* the
limit. The base's 59 packages stay well inside it, which is why only the child
trips — so this is scale-dependent and any overlay adding a large stack will hit
it. Batching the artifact list would fix it, but that is a change to the install
path and belongs in its own PR with tests rather than in a template change.

#### Host requirement worth flagging separately

On Ubuntu 22.04 the resize path fails outright with
`lsblk: unknown column: START,TYPE` — `internal/image/overlay/resize.go:269` reads
partition start sectors via `lsblk -o PATH,START,TYPE`, and the `START` column
landed in util-linux 2.38. `resize.go:239` checks that `lsblk` *exists* but not that
it supports the column it needs, so the failure surfaces late and cryptically. A
capability check, or an `sfdisk -d` fallback, would make overlay resize work on
22.04 too. The guard itself is sound and I did not weaken it: read from the
published image, the noble cloud layout puts root at sector 2099200 and the ESP at
10240, so root *is* the last partition and growing it is safe.

DebalGhosh:
### Resolution PRs raised for the two defects this branch surfaced

Both were found while building the overlay robotics chain here, and both are tool
bugs rather than template problems — so they are fixed in their own PRs off `main`
rather than folded into this one:

| PR | Fixes | Effect on this branch |
|---|---|---|
| **#804** — `fix(overlay): read partition starts with sfdisk when lsblk lacks START` | Overlay resize aborts on Ubuntu 22.04 with `lsblk: unknown column: START,TYPE`, because the `START` column only exists from util-linux 2.38 and 22.04 ships 2.37.2. Falls back to `sfdisk -d`, and adds `sfdisk` to the up-front tool probe so the failure can no longer surface late. | Lets `ubuntu24-x86_64-robotics-hw-overlay-qcow2.yml` resize on 22.04 hosts. It already builds and boots on 24.04. |
| **#805** — `fix(overlay): batch dpkg -i so large overlays stay under the arg limit` | A large overlay fails with `argument list too long` before dpkg starts: all artifact paths go into one `dpkg -i`, and because commands run via `bash -c` the limit is `MAX_ARG_STRLEN` (128 KiB), not `ARG_MAX`. 2004 ROS 2 artifacts produce a ~142 KiB command line. Batches the list inside the existing multi-pass loop, so Pre-Depends convergence is unchanged. | **Unblocks `ubuntu24-x86_64-robotics-jazzy-overlay-extends.yml`**, the last thing standing between that template and a finished artifact. |

Neither changes any template, so nothing here needs rebasing on them. The two
templates in this PR are unchanged and still stand on their own:
`resolve --full` passes all ten structural assertions, and the overlay base builds
and boots (1.67 GB qcow2, 29 targets reached, 0 failed units — see the
[build report](https://github.com/open-edge-platform/image-composer-tool/pull/796#issuecomment-5201096305)).

Both PRs are deliberately narrow: #804 preserves the fail-closed partition guard
exactly (a missing or unparseable start still aborts rather than defaulting to 0),
and #805 leaves the rpm backend alone because rpm installs its artifact set as a
single transaction and chunking it would change semantics rather than just command
length.

Merge order does not matter, and neither blocks this PR.

DebalGhosh:
### The kernel.version fix has been split back out into #806

The `fix(config): drop stale kernel.version pin from the ubuntu24 OS defaults`
commit has been **removed from this branch** and now stands alone as **#806**. It is
an urgent build-blocking fix and should not wait on this larger refactor.

This branch is now 13 commits and no longer touches
`config/osv/ubuntu/ubuntu24/imageconfigs/defaultconfigs/`. Everything else is
unchanged: the per-distribution reorg, the `extends` decompositions, the
overlay+extends robotics chain, and the docs. Re-verified after the drop — build,
`go vet` and the full test suite pass, and every template validates except the
pre-existing `ubuntu24-x86_64-fde-raw.yml`.

**Merge-order note:** #806 and this PR both edit
`docs/user-guide/release-notes.md`, each adding its own bullet to a different
section (#806 under **Fixed**, this PR under **New** and **Fixed**). Whichever
merges second may need a trivial "keep both bullets" resolution — no template or
config overlap.

Related resolution PRs raised from findings on this branch:

| PR | Fixes |
|---|---|
| **#806** | stale `kernel.version: "6.17"` in the ubuntu24 OS defaults (urgent; also repairs `build-ubuntu24-immutable`, `-raw` and `-arm-raw`) |
| **#804** | overlay resize aborting on Ubuntu 22.04 (`lsblk: unknown column: START,TYPE`) |
| **#805** | `dpkg -i` exceeding `MAX_ARG_STRLEN` on large overlays (`argument list too long`) |

DebalGhosh:

---

```
```

|---|---|
| **#806** | `''` ✅ |

---

Fails on **#796 and #806**:

```
  checksum mismatch:
```

| Branch | Result |
|---|---|
| `main` | success |

---

```
```

```
```

three branches.

say the word.

---

#### Summary

| Failure | Cause | Owner |
|---|---|---|

DebalGhosh:

|---|---|---|

```
```

|---|---|---|

#### Current state of #806

```
```

first.

DebalGhosh:
### `build-ubuntu24-immutable` fixed — now green

**Root cause: this PR lost the kernel fix when we extracted it.**

The stale `kernel.version: "6.17"` removal originally lived on this branch. It was pulled out into its own PR (#806) because it is an urgent build-blocker that shouldn't wait on this larger refactor. That left this branch inheriting `main`'s failure:

```
```

`build-ubuntu24-immutable` builds `ubuntu24-x86_64-edge-raw.yml`, which sets no `kernel.version` of its own and inherits `6.17` from `config/osv/ubuntu/ubuntu24/imageconfigs/defaultconfigs/default-raw-x86_64.yml`. The Ubuntu 24.04 archive never carried a `6.17` kernel, so the build fails at package resolution in ~2 minutes.

**Fix:** #806's branch is merged into this one — see the merge commit.

I used a **merge**, not a cherry-pick, deliberately: it records that this PR *contains* #806's commit, so whichever of the two lands second dedupes on shared history instead of conflicting on a duplicate commit. A cherry-pick would have recreated exactly the duplication the extraction removed.

**#806 is unchanged** and remains the standalone urgent fix. It should still merge on its own timeline; this PR just no longer depends on that happening first.

<details><summary>Conflict resolution and verification</summary>

One conflict, `docs/user-guide/release-notes.md` — both branches append to the **Fixed** section. Resolved by keeping **both** sets of bullets.

Worth noting: `image-templates/ubuntu24/robotics-demo-ubuntu24-x86_64.yml` **auto-merged across the rename**. This PR moves it into `ubuntu24/` while #806 edits it at the old flat path; git's rename detection carried the edit over. That was the one interaction I checked explicitly.

Verified locally before pushing:

| Check | Result |
|---|---|
| `grep -rn 'version: "6.17"'` across `config/osv/ubuntu/ubuntu24/` + `image-templates/` | no matches |
| `resolve --full` on `ubuntu24-x86_64-edge-raw.yml` | `kernel.version` = `''`, metapackage list unchanged |
| Conflict markers remaining | 0 |
| Template validation, all templates | only the pre-existing `ubuntu24-x86_64-fde-raw.yml` failure (absolute `passphraseFile` path, unrelated) |
| `go build` / `go vet` / `go test` / `gofmt -l` | clean, 0 test failures |

</details>

<details><summary>CI on the new head: 37 success, 2 skipped, 1 neutral, 0 failures</summary>

`build-ubuntu24-immutable` — **success** (11 min).

`build-elxr-edge-26-04-immutable` failed on the first attempt and **passed on re-run**. It was not caused by this merge, and I confirmed that rather than assuming it:

- The merge commit changes **6 files** vs its parent — 4 ubuntu24 OS defaults, the release notes, and `robotics-demo`. **Zero** eLxr files.
- The failure was upstream repo signature verification, not kernel resolution: `openpgp: invalid signature: RSA verification failure` → `all 4 repositories failed to parse`. It aborted in 215 s during package download, before any image work.
- The same job passed on the five preceding commits of this branch (334 s, 303 s, 327 s, 306 s, 347 s) and on recent `main` heads.

</details>

DebalGhosh:
### Follow-up stacked on this branch: #808 (template CI gate)

While looking at whether the new templates needed build workflows, I found that **nothing in CI checked any template at all** — no workflow ran `validate` or `resolve`, and no Go test resolves a real template ([`validate_test.go#L28`](https://github.com/open-edge-platform/image-composer-tool/blob/main/internal/config/validate/validate_test.go#L28) reads one azl3 file, schema-only). That gap is most relevant right after this PR, since moving a template breaks its `extends:` chain and nothing would have said so.

#808 adds one workflow that runs `validate` + `resolve --full` over all **64** templates in ~5 s, offline. It is **stacked on this branch**, so it does not touch this PR or re-trigger its fan-out. Retarget to `main` after this merges.

Notably it did **not** add build workflows for the 6 new templates, on measured grounds: the dangerous composition failures are silent (giving an overlay child its own `baseline:` collapses its resolved packages 19 → 8 and drops every Intel package, while `validate` exits 0 and the build succeeds — `merge.go:154-157`), six build jobs would add ~40–60 min to an already ~154-runner-minute fan-out, and one of the templates cannot build until #805 lands.

It also carries two small fixes needed to make the gate green on day one, both pre-existing on `main`: `ubuntu24-x86_64-fde-raw.yml` pointed `passphraseFile` at `/home/user/test-fde.txt` (the only template failing `validate` on a fresh clone), and `image-templates/README.md` claimed 24 of 64 templates are built per PR when the real number is 21.

yktyagi (approved):

cmd/image-composer-tool/build_signal_integration_test.go
```diff
@@ -51,7 +51,7 @@ func TestBuildSIGTERMCleansUpMountsAndLoops(t *testing.T) {
 	// Pick a small user template we know builds against Ubuntu 24. The exact
 	// template doesn't matter — we only need the build to progress far enough
 	// to mount /proc/sys/dev into the chroot before we interrupt it.
-	templatePath := filepath.Join(repoRoot, "image-templates", "ubuntu24-x86_64-edge-raw.yml")
```
srmungar, line 54: please resolve merge conflicts.

yktyagi:
Will be creating a new PR

PR #801, opened 2026-08-05
Implement POST /templates/validate with field-level errors

Shruti-Raj-Vansh-Singh:
## What

Replaces the `POST /templates/validate` 501 stub with a real implementation that returns **one issue per problem, each with the offending field's path** — so the Advanced tab can show inline, in-place errors.

```json
POST /templates/validate  { "yaml": "<template>" }
200 { "valid": false, "errors": [
  { "path": "image.name",       "message": "does not match pattern ...", "severity": "error" },
  { "path": "target.os",        "message": "value must be one of ...",   "severity": "error" },
  { "path": "target.imageType", "message": "value must be one of ...",   "severity": "error" },
  { "path": "target",           "message": "missing properties: 'dist', 'arch'", "severity": "error" }
]}
```

## Stacked on #800

Based on `advanced-mode-spec-stubs` (PR #800), which added the spec + generated `ValidationResponse`/`ValidationIssue` types this builds on. **Review/merge #800 first**; this diff shows only PR 2's changes. I'll rebase onto `main` once #800 lands.

## How

- **`internal/config/validate/issues.go`** — new `ValidateUserTemplateIssues`, added *alongside* the existing error-returning validators so CLI callers are unchanged. On a schema failure it unwraps the underlying `*jsonschema.ValidationError` with `errors.As` and walks the nested `Causes`, emitting the **leaf** errors (the most specific ones) with `InstanceLocation` converted to a dotted field path. **No string-parsing of the formatted message** — that was called out as fragile. It then runs the same post-schema semantic checks (auto-expand, FDE) the build runs.
- **`internal/api/service/validate.go`** — `ValidateTemplate` converts the YAML the same way the build does (`yaml → interface{} → json`, with the same string-safety limits) and returns a structured result.
- **`internal/api/mapping.go` / `handlers.go`** — map the result to `ValidationResponse`; the handler encodes it.

## The key correctness property

The verdict here **mirrors what building the same YAML would reach**. A `{yaml}` build writes the template to disk and runs ICT's `LoadAndMergeTemplate`, which validates it as a *user (minimal) template* — the same subschema and the same semantic checks this endpoint runs. So a template that validates here builds without a validation failure, and one that fails here fails the build for the same reason, only reported field-by-field.

## Contract: a failed validation is a successful call

- Invalid template → **200** with `valid: false` + issues (not a 4xx).
- Malformed YAML → 200 `valid: false` with a single parse issue (reported, not thrown).
- **4xx only** for a malformed request envelope or an empty template.

## Tests

- `internal/config/validate/issues_test.go` — multi-field-error unwrapping (asserts distinct paths `image.name`, `target.os`, `target.imageType`, `target`), the valid case, missing-required-top-level (document-level issue), the FDE semantic check, and `pointerToPath` (incl. `~0`/`~1` escapes).
- `internal/api/service/validate_test.go` — valid/invalid result semantics, malformed-YAML-is-invalid-not-error, empty-is-400.
- `internal/api/api_test.go` — handler returns 200 (not 4xx) for invalid templates with usable paths; 400 on bad JSON envelope. Removed `validate` from the 501-stub test (now implemented).

## Verification

- `go generate ./internal/api/http` idempotent; `go build`, `go vet`, `go test` — pass
- gofmt clean; golangci-lint v1.64.2 — 0 issues
- Overall coverage 72.6% (gate 69.9%); new code well-covered
- `npm run build` — pass

### DIFF
--- a/internal/api/api_test.go
+++ b/internal/api/api_test.go
@@ -262,7 +262,7 @@ func TestHandleValidateTemplateValid(t *testing.T) {
 	if err := json.Unmarshal(rr.Body.Bytes(), &resp); err != nil {
 		t.Fatalf("decode: %v", err)
 	}
-	if resp.Valid == nil || !*resp.Valid {
+	if !resp.Valid {
 		t.Errorf("valid = %v, want true (body: %s)", resp.Valid, rr.Body)
 	}
 	if resp.Errors != nil && len(*resp.Errors) != 0 {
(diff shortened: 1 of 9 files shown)


arodage (approved):
LGTM

arodage (approved):
LGTM

PR #803, opened 2026-08-06
Enable Advanced tab with read-only template preview

Shruti-Raj-Vansh-Singh:
Enables the previously-disabled **Advanced** tab and adds its first slice: a read-only preview of the resolved template.

## What this does
- Adds `AdvancedPage`, which reuses the Basic tab's cascading selection (shared Zustand store) to resolve a pre-authored template via `POST /templates/compose`, then renders that template's YAML **read-only**.
- Flips the Advanced tab from `enabled: false` to enabled in `App.tsx` and renders the new page.

## Scope
Frontend only — no API, contract, or backend change. This is the tab shell + preview; editing, validation, package selection, and export arrive in later PRs.

## Verification
- `cd web && npm run build` (tsc typecheck + vite bundle) passes; the new component is present in the production bundle.
- Ran the server locally and confirmed `POST /templates/compose` returns the full template YAML the preview renders.


arodage (approved):
LGTM

PR #804, opened 2026-08-06
fix(overlay): read partition starts with sfdisk when lsblk lacks START

DebalGhosh:
#### Merge Checklist  

- [x] Ready to merge

#### Description 

Overlay resize fails outright on Ubuntu 22.04:

```
overlay resize: failed to read partition layout of /dev/loop25:
failed to execute command with output "lsblk: unknown column: START,TYPE"
```

`assertRootIsLastPartition` reads each partition's start offset with
`lsblk -o PATH,START,TYPE` to confirm the root partition is last before letting
`growpart` extend it. **The `START` column only exists from util-linux 2.38 onward**;
Ubuntu 22.04 ships 2.37.2, so the command fails and takes the whole build with it —
*after* the baseline has already been downloaded, resolved and preflighted.

Compounding it, `checkResizeToolsAvailable` probes that `lsblk` **exists** but not
that it supports the column the guard needs, so the failure surfaces late and its
message points at the loop device rather than at the host's util-linux version.

Hit while building an overlay template on a 22.04 host; the repo's README
recommends Ubuntu 24.04, but the failure mode gives no hint that the host version
is the problem.

#### The fix

Fall back to `sfdisk -d`, which reports the same offsets and has carried its dump
format far longer. `sfdisk` is already on the shell allowlist.

- **Only** an unknown-column failure is retried this way. Any other `lsblk` error
  (missing device, permissions) still surfaces as-is rather than being masked by a
  second tool.
- `sfdisk` joins the up-front required-tools probe, so a host missing it fails early
  with the existing actionable message instead of part-way through the guard.

**Units matter here**, since mixing them would be a correctness bug: both sources
report **sectors**. `lsblk`'s `START` is a sector offset even under `-b` — verified
against one disk, where `lsblk` and `sfdisk` both report `2099200` for a noble cloud
image's root. The guard only compares offsets against each other, so a consistent
unit suffices.

#### The safety property is unchanged

`parseSfdiskPartitionStarts` fails closed exactly as the lsblk parser does: a
partition line with a missing or unparseable start **aborts** the guard rather than
defaulting to `0`. A silent `0` would make a later partition look like it precedes
root and could let `growpart` run into it. A dump with no partition lines is an
error, never an empty map.

None. `sfdisk` was already allowlisted in `internal/utils/shell/shell.go`.

`go build ./cmd/... ./internal/...`, `go vet`, `go test ./cmd/... ./internal/...`
(0 failures) and `gofmt -l` all clean.

New unit tests, using the existing `resizeExec` seam so no loop device or root is
needed:

| Test | Asserts |
|---|---|
| `TestPartitionStarts_FallsBackToSfdiskWhenStartColumnUnsupported` | the fallback fires on an unknown-column error and yields the same offsets |
| `TestPartitionStarts_OtherLsblkErrorDoesNotFallBack` | an unrelated lsblk failure surfaces; sfdisk must **not** run |
| `TestAssertRootIsLastPartition_AcceptsViaSfdiskFallback` | a genuinely-last root still passes through the fallback |
| `TestAssertRootIsLastPartition_RejectsNonLastRootViaSfdiskFallback` | a non-last root is still **rejected** — the fallback is a different data source, not a relaxed rule |
| `TestParseSfdiskPartitionStarts` | header skipping (incl. the `device:` line, which has a colon but no `start=`), and fail-closed on unparseable/absent starts |

The real-world sfdisk output the parser was written against was captured from an
actual overlay artifact rather than invented:

```
label: gpt
sector-size: 512

/dev/nbd0p1 : start=     2099200, size=    48232415, type=0FC63DAF-…
/dev/nbd0p15 : start=       10240, size=      217088, type=C12A7328-…
```

### DIFF
--- a/internal/image/overlay/resize.go
+++ b/internal/image/overlay/resize.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math"
 	"os"
+	"strconv"
 	"strings"
 	"unicode"
 
@@ -234,9 +235,12 @@ func resizeToolsForFS(fsType, table string) []string {
 // checkResizeToolsAvailable verifies every host tool the resize sequence needs is
 // present on the build host before any disk mutation, so a missing dependency
 // fails with a clear, actionable message up front rather than mid-sequence. The
-// partition-inspection guard also relies on lsblk, so it is included.
+// partition-inspection guard also relies on lsblk, and on sfdisk as its fallback
+// when lsblk predates the START column (util-linux < 2.38), so both are included:
+// a host missing sfdisk would otherwise only discover it after the guard had
+// already committed to the fallback path.
 func checkResizeToolsAvailable(layout *Layout) error {
-	tools := append([]string{"lsblk"}, resizeToolsForFS(layout.RootFSType, layout.PartitionTable)...)
+	tools := append([]string{"lsblk", "sfdisk"}, resizeToolsForFS(layout.RootFSType, layout.PartitionTable)...)
 	var missing []string
 	for _, t := range tools {
 		ok, err := resizeToolExists(t)
@@ -251,7 +255,7 @@ func checkResizeToolsAvailable(layout *Layout) error {
 		return fmt.Errorf(
 			"overlay resize: required tool(s) not found on the build host: %s; "+
 				"install them (growpart is in cloud-guest-utils, sgdisk in gdisk, "+
-				"resize2fs in e2fsprogs, xfs_growfs in xfsprogs, losetup/partx in util-linux) "+
+				"resize2fs in e2fsprogs, xfs_growfs in xfsprogs, losetup/partx/lsblk/sfdisk in util-linux) "+
 				"or remove/lower disk.size (optionally also set overlayPolicy.allowDiskResize: false) "+
 				"to keep the baseline layout and skip the resize",
 			strings.Join(missing, ", "))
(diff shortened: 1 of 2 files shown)


DebalGhosh:
### Note on the failing `Unit Tests & Coverage Gate`

The failure is **`internal/api: TestHandleStartBuildAccepted`** — a package this PR
does not touch. This PR changes only `internal/image/overlay/resize.go` and its test
file.

Evidence it is not caused by this change:

- The test passes locally, including `go test ./internal/api/ -run TestHandleStartBuildAccepted -count=20` (20/20).
- **#805 passed the identical gate** on the same base with a comparably unrelated
  overlay-only change.
- Overall coverage came out at **70.1%** against a 70.0% threshold, so the gate's
  coverage arithmetic is satisfied; the job fails purely on that one test result.

Reads as a CI-environment flake in `internal/api` rather than a regression here. I
have not tried to fix it in this PR, since doing so would mean touching an unrelated
package inside a narrowly-scoped overlay fix — happy to raise it separately if
maintainers would like.

`build-ubuntu24-immutable` also fails, and that one is a real pre-existing bug —
fixed in **#806** (stale `kernel.version: "6.17"` in the ubuntu24 OS defaults,
reached via `ubuntu24-x86_64-edge-raw.yml`).

DebalGhosh:

---

```
```

|---|---|
| **#806** | `''` ✅ |

---

Fails on **#796 and #806**:

```
  checksum mismatch:
```

| Branch | Result |
|---|---|
| `main` | success |

---

```
```

```
```

three branches.

say the word.

---

#### Summary

| Failure | Cause | Owner |
|---|---|---|

DebalGhosh:

|---|---|---|

```
```

|---|---|---|

#### Current state of #806

```
```

first.

PR #808, opened 2026-08-06
ci(templates): gate every image template on validate and resolve

DebalGhosh:
## Problem

Nothing in CI checked the image templates.

No workflow ran `validate` or `resolve`, and no Go test resolves a real template — [`internal/config/validate/validate_test.go#L28`](https://github.com/open-edge-platform/image-composer-tool/blob/main/internal/config/validate/validate_test.go#L28) reads a single azl3 file and checks it against the JSON schema only. The `extends`/overlay merge in `internal/config/merge.go` is well covered by unit tests, but only against synthetic in-test YAML. Nothing asserted that the 64 templates *in this repo* satisfy it.

That gap matters most right after #796, which moved every template into a per-distribution subdirectory. `extends:` parents resolve relative to the child and `..` is rejected, so moving a parent breaks the chain — and 43 of the 64 templates are never built on a PR either, so nothing would have said so.

## What this adds

One workflow, `Template Gate`, running `scripts/template_gate.sh` over every path in `git ls-files 'image-templates/*.yml'`:

| | |
|---|---|
| Coverage | **64 of 64** templates, up from 21 built on a PR |
| Runtime | **~5 s** (measured from a fresh clone) |
| Needs | Go only — no root, Docker, Earthly, or apt |
| Network | **none** — verified passing under `unshare -rn` |

Because discovery is `git ls-files`, a template added in a later PR is covered the moment it is committed; no workflow to edit. It also ignores untracked scratch files in a working tree.

`resolve --full` is the load-bearing half — it folds the extends chain *and* the OS defaults. Verified by breaking things on purpose:

| Injected regression | Result |
|---|---|
| `extends:` parent moved away | `extends path not found` → exit 1 |
| `extends: ../other-dist/...` | `extends path escapes child template's directory` → exit 1 |
| OS defaults unreadable | fails as vacuous → exit 1 |
| Run from wrong directory | fails on the missing `./config` guard → exit 1 |

**Exit status alone is not trusted.** Run from the wrong directory the tool cannot find `./config`, warns `Could not load default configuration`, and **still exits 0** — a naive gate would pass while checking nothing. The script fails on that string, and refuses to run without `./config` or with an empty template list.

## Deliberately *not* added: a build workflow per template

This started as "add workflows for the new templates". I measured it, and build jobs are the wrong instrument:

1. **The dangerous failures are silent.** Give an overlay child its own `baseline:` and its resolved package list collapses **19 → 8**, dropping every Intel hardware package (`intel-oneapi-runtime-compilers`, `libze-intel-gpu1`, `intel-level-zero-npu`) — while `validate` exits 0 and the build *succeeds*. Cause is the unconditional package replacement under `IsOverlayMode()` at `internal/config/merge.go:154-157`. Building all 64 templates would not catch this; comparing resolved output does.
2. **Cost.** One PR head already burns **~154 runner-minutes across 40 checks** with no path filters (measured on `ed945e2b`), critical path 20m18s. Six build jobs add ~40–60 min and six more chances of red. This gate adds ~1–2 min and 0 to the critical path.
3. **One of them cannot build yet.** `ubuntu24-x86_64-robotics-jazzy-overlay-extends.yml` hits `argument list too long` — the overlay installer passes all 2004 artifact paths in one `dpkg -i` (~141932 bytes vs `MAX_ARG_STRLEN` 131072). Fixed separately in #805; a build job here would be guaranteed red.

Follow-up work worth doing, deliberately out of scope: golden assertions on resolved output (the silent modes above), an `additionalFiles` existence check (currently would fail on one genuine pre-existing break in `debian13-x86_64-desktop-virtualization-iso.yml`), and path filters + concurrency on the 21 heavy build workflows — that last one touches branch protection and belongs to a maintainer.

## Drive-by fixes needed to make it green

- **`ubuntu24-x86_64-fde-raw.yml`** — `passphraseFile` was `/home/user/test-fde.txt`, an absolute path from a developer's home directory. This was the **only** template failing `validate` on a fresh clone, and it fails on `main` today. Now points at `fde-passphrase.example.txt` committed beside the template; a relative `passphraseFile` resolves against the template's own directory ([`config.go:908-909`](https://github.com/open-edge-platform/image-composer-tool/blob/main/internal/config/config.go#L908)). The value is a placeholder, and the tool redacts it in `resolve` output. Commenting the field out is not an option — the schema requires it whenever `fde.enabled` is true.
- **`image-templates/README.md`** — said "24 of the 64 templates are built … on every pull request". The real number is **21**: there are 23 `build-*.yml` workflows and two are deliberately disabled. Also corrected the claim that the rest "are validated by schema" — until this PR, nothing validated them.

## Verification

- `scripts/template_gate.sh` — **64/64 pass in 5.1 s from a pristine clone**, and again under `unshare -rn`
- All four negative tests above fail correctly with exit 1
- `shellcheck scripts/template_gate.sh` — clean
- `go build` / `go vet` / `go test ./cmd/... ./internal/...` — **0 failures**; `gofmt -l` clean
- Workflow follows `go-lint.yml`: actions pinned to the same full commit SHAs, `permissions: contents: read`, `persist-credentials: false`, concurrency group. No `paths:` filter, so it is safe to make a required check (a filtered check sits pending forever on unrelated PRs).

## Base branch

Stacked on `refactor/image-templates-categorize-and-compose` (#796) rather than `main`, because the reorg is what makes the extends-chain check worth having. Retarget to `main` after #796 merges. The `fde-raw` fix is independent of #796 and could be cherry-picked out if that is preferred.


DebalGhosh:
### Note on why CI looks quiet here

Only the 4 gitleaks checks ran on this PR (all green — the example passphrase file is clean). **The Template Gate itself has not run in CI yet, and cannot until this lands on `main`.** Two separate reasons, both worth stating so nobody reads the empty check list as a pass:

1. **The `pull_request` trigger is scoped to `branches: [main]`,** matching every `build-*.yml` in the repo. This PR targets `refactor/image-templates-categorize-and-compose`, not `main`, so no build workflow fires — mine included. `gitleaks` ran only because it uses an unqualified `on: [pull_request, ...]` with no branch filter. This is existing repo-wide behavior, not something specific to this workflow.
2. **`workflow_dispatch` is not available either.** GitHub only registers a workflow for manual dispatch once the file exists on the default branch; `gh workflow run template-gate.yml` returns `could not find any workflows named template-gate.yml`, and the workflow is absent from the 47 registered ones.

So the evidence for this gate is local, and I ran it the way CI will rather than the way that is convenient:

```
$ git clone --branch ci/template-static-gate … /tmp/cisim   # pristine clone, no working-tree state
$ go build -buildmode=pie -ldflags "-s -w" -o ./build/image-composer-tool ./cmd/image-composer-tool
$ ./scripts/template_gate.sh
Checking 64 templates
All 64 templates passed validate and resolve --full.
real 0m5.146s

$ unshare -rn ./scripts/template_gate.sh      # no network namespace
All 64 templates passed validate and resolve --full.   → exit 0
```

The fresh clone matters: it proves the result does not depend on a stale local binary or on the untracked scratch files in my working tree — and that the `fde-raw` template, which fails `validate` on `main` today, passes here.

I would rather flag this than claim green checks I do not have. **First real CI execution will be on the first PR that targets `main`** — either this one after retargeting, or the follow-up once #796 merges. If you would prefer to see it run before review, I can retarget this PR to `main`; the cost is that the diff then also contains #796's commits until that merges.

DebalGhosh:
Superseded by **#809**, which is the same change targeting `main`.

Closing because this PR could never demonstrate itself in CI. Every `build-*.yml` scopes `pull_request` to `branches: [main]`, so a PR based on `refactor/image-templates-categorize-and-compose` triggers no workflow — including the one it adds. Only `gitleaks` ran here (4/4 green), because it uses a bare `on: [pull_request, ...]` with no branch filter. `workflow_dispatch` was no help either: GitHub does not register a workflow for manual dispatch until the file is on the default branch.

Retargeting this PR to `main` was not viable: the head branch is stacked on #796, so the diff became 127 files (+2184/−585) instead of 6 — it swallowed the whole reorg. A PR's head branch cannot be changed after creation, so #809 was opened from a clean `main`-based branch instead.

**#809 is verified green in real CI** — `Validate and resolve all image templates` passed in 54 s, logging `Checking 60 templates` → `All 60 templates passed validate and resolve --full`, with no `Could not load default configuration`, so it did real work rather than passing vacuously.

Nothing is lost by closing this. The only difference is layout — 60 flat templates in #809 vs 64 in per-distribution subdirectories here — and after #796 merges, #809 needs a trivial rebase to pick up the subdirectory paths. The `fde-raw` fix is independent of #796 either way: it is the sole gate failure on `main` too.

PR #809, opened 2026-08-06
ci(templates): gate every image template on validate and resolve

DebalGhosh:
## Problem

Nothing in CI checked the image templates.

No workflow ran `validate` or `resolve`, and no Go test resolves a real template — [`internal/config/validate/validate_test.go#L28`](https://github.com/open-edge-platform/image-composer-tool/blob/main/internal/config/validate/validate_test.go#L28) reads a single azl3 file and checks it against the JSON schema only. The `extends`/overlay merge in `internal/config/merge.go` is well covered by unit tests, but only against synthetic in-test YAML. Nothing asserted that the templates *in this repo* satisfy it — and most of them are never built on a PR either.

## What this adds

One workflow, `Template Gate`, running `scripts/template_gate.sh` over every path in `git ls-files 'image-templates/*.yml'`:

| | |
|---|---|
| Coverage | **60 of 60** templates |
| Runtime | **4.7 s** (measured from a pristine clone) |
| Needs | Go only — no root, Docker, Earthly, or apt |
| Network | **none** — verified passing under `unshare -rn` |

Because discovery is `git ls-files`, a template added in a later PR is covered the moment it is committed; no workflow to edit. It also ignores untracked scratch files in a working tree.

`resolve --full` is the load-bearing half — it folds the extends chain *and* the OS defaults. Verified by breaking things on purpose:

| Injected regression | Result |
|---|---|
| `extends:` parent moved away | `extends path not found` → exit 1 |
| `extends: ../other-dir/...` | `extends path escapes child template's directory` → exit 1 |
| OS defaults unreadable | fails as vacuous → exit 1 |
| Run from wrong directory | fails on the missing `./config` guard → exit 1 |

**Exit status alone is not trusted.** Run from the wrong directory the tool cannot find `./config`, warns `Could not load default configuration`, and **still exits 0** — a naive gate would pass while checking nothing. The script fails on that string, and refuses to run without `./config` or with an empty template list.

## Deliberately *not* added: a build workflow per template

This began as "add build workflows for the templates that lack them". Measured, build jobs are the wrong instrument:

1. **The dangerous failures are silent.** Giving an overlay child its own `baseline:` collapses its resolved package list and drops packages the parent contributed — while `validate` exits 0 and the build *succeeds*. Cause is the unconditional package replacement under `IsOverlayMode()` at `internal/config/merge.go:154-157`. Building every template would not catch this; comparing resolved output does.
2. **Cost.** One PR head already burns **~154 runner-minutes across 40 checks** with no path filters, critical path 20m18s. Each new build job adds minutes and another chance of a red check. This gate adds ~1–2 min and 0 to the critical path.

Follow-up work, deliberately out of scope: golden assertions on resolved output (the silent modes above), an `additionalFiles` existence check (currently would fail on one genuine pre-existing break in `debian13-x86_64-desktop-virtualization-iso.yml`), and path filters + concurrency on the heavy build workflows — that last one interacts with branch protection and belongs to a maintainer.

## Drive-by fix needed to make it green

`ubuntu24-x86_64-fde-raw.yml` had `passphraseFile: "/home/user/test-fde.txt"` — an absolute path from a developer's home directory. This is the **only** template that fails `validate` on a fresh clone of `main` today:

```
Error: validation failed: failed to read systemConfig.fde.passphraseFile
"/home/user/test-fde.txt": … no such file or directory
```

It now points at `fde-passphrase.example.txt`, committed beside the template; a relative `passphraseFile` resolves against the template's own directory ([`config.go:908-909`](https://github.com/open-edge-platform/image-composer-tool/blob/main/internal/config/config.go#L908)). The value is a placeholder, not a usable secret, and the tool redacts it in `resolve` output. Commenting the field out is not an option — the schema requires it whenever `fde.enabled` is true.

## Verification

- `scripts/template_gate.sh` — **60/60 pass in 4.7 s from a pristine clone**, and again under `unshare -rn`
- All four negative tests above fail correctly with exit 1
- `shellcheck scripts/template_gate.sh` — clean
- `go build` / `go vet` / `gofmt -l` — clean
- Workflow follows `go-lint.yml`: actions pinned to the same full commit SHAs, `permissions: contents: read`, `persist-credentials: false`, concurrency group. No `paths:` filter, so it is safe to make a required check (a filtered check sits pending forever on unrelated PRs).

## Relationship to #808

#808 is the same change stacked on #796 (`refactor/image-templates-categorize-and-compose`), where templates live in per-distribution subdirectories. Because every `build-*.yml` scopes `pull_request` to `branches: [main]`, no workflow — including this one — can run on a PR that targets a non-`main` branch, so #808 could never demonstrate itself in CI. **This PR targets `main` so the gate actually executes.**

The two differ only in layout: 60 templates flat here vs 64 in subdirectories there, and the `fde-raw` fix at the flat path. Merge whichever suits the order you want; the other becomes a small rebase. Worth noting the `fde-raw` fix is independent of #796 — it is the sole gate failure on `main` as well.


DebalGhosh:
### The gate works: verified green in real CI

`Validate and resolve all image templates` — **success in 54 s**:

```
Checking 60 templates
All 60 templates passed validate and resolve --full.
```

Tool time was 3.4 s of that; the rest is checkout plus `go build`. No `Could not load default configuration` in the log, which matters — that string is how the script detects a run that found no OS defaults and would otherwise report a vacuous pass.

### `build-ubuntu24-immutable` is red, and it is not this PR

It fails in 3m42s with:

```
```

That is inherited from `main`, not caused here:

- This PR touches **zero** files under `config/osv/` (`git diff --name-only origin/main..HEAD | grep -c config/osv` → `0`).
- `main` still pins `kernel.version: "6.17"` in `config/osv/ubuntu/ubuntu24/imageconfigs/defaultconfigs/default-raw-x86_64.yml:81`, and Ubuntu noble no longer ships 6.17.
- The same job is **0/4 on recent `main` heads**, and it fails identically on #804 and #805 — every PR branched off `main` inherits it.

**#806 is the fix** and is green, awaiting review. Once it lands, this check goes green here with no change to this PR.

PR #835, opened 2026-08-17
feat(builder): add builder image targets and fail-fast toolchain preflight

tejaswin98:
Adds builder container image support and a fail-fast preflight check for the builder toolchain.

### Summary
- add `builder-image` and `builder-image-push` targets to build and publish the builder container image
- add `builder-image-preflight` to validate required shell allowlist tools before runtime
- add a dedicated builder container definition with the ICT toolchain bundled
- add a preflight script that parses the shell command allowlist and fails if required tools are missing
- add a GitHub Actions workflow to run the builder image preflight on relevant changes

### Why
This makes the builder image reproducible and catches missing toolchain components early in CI instead of failing later at runtime.

No new third-party Go dependencies were introduced.

###How Has This Been Tested?

- `earthly +builder-image-preflight`
- `earthly +builder-image`
- `earthly +test-quick`
- `earthly +lint`


arodage (approved):
LGTM

PR #843, opened 2026-08-20
docs: add ADR for Advanced mode template modification via extends

arodage:
#### Merge Checklist  

- [ ] The changes in the PR have been built and tested
- [ ] Ready to merge

#### Description 

Adds an ADR proposing how the web UI's Advanced mode should apply user
modifications to a curated template before building.

**Decision:** the server generates a small delta template that `extends` the
curated one and builds that. The Review step displays `resolve --full` of the
same delta, so the template a user reviewed and the template that gets built
are the same file going through the same merge engine and cannot drift apart.
Advanced mode becomes a front end for the template inheritance ICT already
ships (`adr-template-extends.md`) rather than a new override mechanism, so each
additional editable attribute is one more field in the delta instead of a new
CLI flag, wire field, and validation path.

The ADR also records the two constraints that shape the design, both of which
are easy to rediscover the hard way:

- The `extends` containment guard (`resolveExtendsParentPath`) requires the
  parent to resolve at or below the child's directory and rejects `..`. Since
  curated templates live in `TemplatesDir`, a generated delta can only live
  there or in an ancestor, **not** in the per-build work directory.
- `targetsMatch` compares `imageType`, so a delta cannot change image type.
  That is the same underlying reason a RAW/ISO override cannot work at all:
  `target.imageType` selects which per-type default config layer is merged in,
  and `Target`/`Disk` are wholesale replaces with no unset sentinel. Supporting
  multiple image types per combination needs separate curated templates per
  type, so the control is displayed but disabled for now.

This is documentation only. No functional change, no new dependencies.

Not applicable: the change adds a single Markdown file under
`docs/architecture-decision-record/` and touches no code, build, or CI
configuration. Verified the rendered Markdown and that its relative links to
`adr-template-extends.md`, `adr-image-extension.md`, and
`adr-web-ui-tech-stack.md` resolve.


docs/architecture-decision-record/adr-web-ui-advanced-mode-extends.md
```diff
@@ -0,0 +1,234 @@
...
+
+The merge system already supports exactly the operation Advanced mode needs.
+`extends` lets a child template restate a few fields and inherit everything
+else, with per-attribute merge strategies (override, additive, merge-by-key)
+that are already specified and tested.
+
+### Key Constraint: where a generated delta may live
+
+`resolveExtendsParentPath` (`internal/config/merge.go`) requires the parent to
+resolve at or below the **child's** directory and rejects `..`. Curated
```
srmungar, line 54: Storing the advanced templates on the server-side may not scale for the case that the server for the enterprise. Advanced templates can be saved to user's local directory and as needed user can pass it back to the server to build the image. However, the basic templates needs to be stored on the server side as this is 1) coming from Intel but not from user 2) These are to be referenced/used by all the users.

arodage: This section only states the extends containment guard: resolveExtendsParentPath requires a parent to resolve at or below the child's directory, which is why a delta can't live in the per-build work directory. It doesn't decide anything about storage or persistence. 
Advanced mode deltas aren't stored server-side. They're ephemeral, written, resolved, built, then deleted, including on failed and cancelled builds. There's no per-user template accumulation, so there's nothing that grows with user count.

srmungar: though it is temporary store of the delta template, it is good to maintain a separate folder/path instead of inside the server binary/source paths. However, based on discussion it is okay to move forward due to the constraint due to the ICT binary operation.

srmungar (approved):

PR #847, opened 2026-08-24
Use wildcard kernel packages in handheld template

DebalGhosh:
## Change

Update `image-templates/ubuntu24/generic-handheld-os-template.yml` to use wildcard kernel package requests:

- Replace `linux-image-6.18-intel_260427t075939z-r2` with `linux-image-6.18-intel*`.
- Replace `linux-headers-6.18-intel_260427t075939z-r2` with `linux-headers-6.18-intel*`.
- Remove the fixed `kernel.version` value.

This allows the template to select the matching current kernel package revision from the configured repository. No other files are changed.


arodage (approved):
LGTM

PR #869, opened 2026-08-31
feat(webui): editable Disk Layout step in the Advanced tab

Shruti-Raj-Vansh-Singh:
#### Merge Checklist  

- [ ] Ready to merge

#### Description 

Step 3 of the Advanced wizard ("Disk Layout") was a placeholder. This makes it a
working editor for the resolved template's `disk` block, seeded from whichever
template the current selection composes to.

**Frontend only** — no Go, no OpenAPI, no schema changes.

What it edits:

- **Disk name** (schema-required) and **disk size** with a GiB/GB/MiB unit.
- **Partition table** — GPT *and* MBR. The schema allows both; the design
  prototype only offered GPT.
- **Partitions** — add, remove, reorder, resize, set mount point / fsType, and a
  details panel for `id`, `type`, `typeUUID`, `fsLabel`, `mountOptions`, `flags`.
- **`disk.artifacts[]`** — output format (`raw`, `qcow2`, `vhd`, `vhdx`, `vmdk`,
  `vdi`, `tar`) plus optional compression.
- **Reset to template defaults**, and a collapsible preview of the `disk:` block
  the step would contribute.

**Partitions can be edited two ways, switchable per disk.** The schema stores
`start`/`end` and no per-partition size field. Deriving the offsets keeps a
layout contiguous by construction, which is what you want almost always — but it
makes a deliberate gap, or reproducing an existing table offset-for-offset,
impossible to express. So there is one model with two modes:

| Mode | You edit | Derived | Contiguity |
|---|---|---|---|
| **Size-based** (default) | one size per partition | `start`/`end` | guaranteed — never stored, so it cannot be broken |
| **Offset-based** | `start`/`end` as typed | size per partition | yours; gaps and overlaps are **warned** about, not blocked |

Switching converts the layout in place in both directions, and an unedited
layout emits identical YAML either way. Gaps and overlaps are warnings rather
than errors because the builder accepts both — a gap is legal and sometimes
deliberate, and blocking an overlap would be the UI inventing a rule the builder
does not have. Rationale and the alternatives are in
[`adr-web-ui-disk-step.md`](docs/architecture-decision-record/adr-web-ui-disk-step.md).

**Every field in `$defs.Disk` round-trips.** The partition row shows `name`,
`fsLabel`, `fsType`, size, `mountPoint`, `start` and `end` as labelled columns;
`id`, `index`, `type`, `typeUUID`, `mountOptions` and `flags` are editable behind
a per-row details toggle. `Disk.path` and `Disk.selectionPolicy` are read and
re-emitted but deliberately not editable — they select a *target disk* for the
live installer, which is not what an image-composition step configures.
`Disk.maxSize` is shown only when the resolved template already has one, since it
is an overlay-mode ceiling the build otherwise ignores.

**Bug fixed by that audit: `index` was silently dropped.** Neither read nor
written, so a template setting one lost it on the way through the editor. No
template in this repo uses it (the Go field is `*int` with `omitempty`), which is
exactly why it needed a test rather than being caught by the golden round-trip.

**QCOW2 lives on `disk.artifacts[].type`, not on Image Type.** `target.imageType`
(`raw`/`img`/`iso`/`wsl2`) is how the image is built; `disk.artifacts[].type` is
what container it is written to. QCOW2 only exists on the second, which is why
the control is on this step.

**Seeding.** `POST /templates/compose` already returns the fully merged template
(equivalent to `resolve --full`), so the step parses the `disk` block out of that
YAML. The compose *summary* carries only aggregates (`diskSize`,
`partitionCount`, `partitionTable`), not the partition list. This avoids widening
the API contract for data the response already carries.

**Validation mirrors the loader, not just the schema.** Two rules in
`internal/config/validate/validate.go` are enforced outside the JSON schema:
`extendLastPartitionToFillDisk` is rejected for `imageType: iso`, and for `raw`
unless the last partition is the rootfs (`/`). A disk block can be schema-clean
and still be refused before a build starts. Both were found by running this
step's output through `image-composer-tool validate` — not by reading the schema
— and are now reported inline as you edit.

#### The layout reaches the build

The Disk step edited a model nothing consumed. It now goes out as a typed `disk`
object on `POST /templates/compose` and `POST /builds`, and the backend emits it
into the **same generated `extends` delta** the image-name and package overrides
already use ([`adr-web-ui-advanced-mode-extends.md`](docs/architecture-decision-record/adr-web-ui-advanced-mode-extends.md)).
The Review step's "Your changes" view, the resolved template, and the built image
are one file, merged one way.

Two things follow from how `disk` merges, and both are easy to get wrong:

**The merge replaces `disk` wholesale** (`config/merge.go:130`), unlike
`systemConfig` where package lists are unioned. So the override is always the
**complete block, never a diff** — a partial one would silently delete the
parent's whole partition table. That is also why the wire type carries `path` and
`selectionPolicy`, which the step does not edit: they are read from the resolved
template and sent back unchanged, because omitting them would drop them.
`image-templates/ubuntu24/ubuntu24-x86_64-minimal-unattended-iso.yml` is a real
template that sets both.

**An unedited layout is not sent at all.** It round-trips to the parent's own
disk block, so sending it would generate a delta that changes nothing while
making the Review pane report an override.

#### Validation is the only gate

Nothing validates request bodies against the OpenAPI spec at runtime — the
handlers decode straight into the generated structs — so the spec's `pattern`,
`enum`, `maxLength` and `additionalProperties: false` are documentation. The
typed decode drops unknown keys by construction; `service.ValidateDisk` enforces
everything else, transcribed from the builder rather than the template schema.
`POST /builds` validates independently rather than trusting a prior compose,
matching what `ValidateImageName`/`ValidatePackages` already do.

#### Also fixed: Advanced builds were dropping the package picks

`buildReq` was rebuilt from `selection` and listed only `imageName`, so every
Advanced build silently discarded the Packages step's selections from #863. It
now spreads `composeReq` and re-reads only the two debounced fields live. The
disk override would have landed in the same hole.

#### Not in this PR

`disk.path` and `disk.selectionPolicy` are carried and re-emitted but not
editable — they select a *target disk* for the live installer, not an image
layout. The auto-expand rule that requires immutability to be disabled lives in
`systemConfig`, which this step neither reads nor edits; `POST /templates/validate`
catches it.

| Package | Kind | License | Why |
|---|---|---|---|
| `js-yaml` ^4.3.2 | runtime | MIT | Parses the `disk` block out of the merged template the compose endpoint returns, and renders the preview. Client-side template generation will need the emitter too. Production bundle: 322.05 kB (99.76 kB gzip). |
| `@types/js-yaml` ^4.0.9 | dev | MIT | Types for the above. |
| `vitest` ^3.2.7 | dev | MIT | `web/package.json` had no test script and there were no `*.test.*` files under `web/`. Everything worth testing here is pure, so this is vitest alone — no jsdom, no Testing Library. CI does not build or test `web/` today, so no workflow changes were needed. |

**1. The override reaches the image.** Compose with an edited disk returns a
resolved template with `size: 48GiB`, `partitionTableType: mbr` and the added
`/data` partition; the summary row agrees (`48GiB, MBR, 4 partitions`).

**2. An unedited override is a byte-exact no-op** — for all four buildable
combinations, in **both** layout modes, compared on the resolved `disk` block
(12/12). This is the guard against the step quietly changing a template nobody
asked it to change.

**3. 11/11 malformed overrides refused over HTTP** with a field-level message:
bad `fsType`, decimal size, bare-byte offset, path traversal, name traversal,
`tar` on RAW, `bz2`, interior "rest" partition, fill-disk on ISO, missing name.
`POST /builds` refuses the same set independently.

**4. The build path.** `POST /builds` accepted the override, merged it
(`Successfully created merged configuration with ... disk config:
minimal-desktop-ubuntu-ptl-pv`), and archived a resolved template carrying it.

**5. Browser** (Chromium → Vite → API): four rapid edits collapsed into **one**
compose POST (debounce), an untouched step sent no override at all, and the
Review step's "Your changes" and "Resolved" views both show the edited block. No
console errors.

**6. Static + unit gates.** `go build`, `go vet`, `gofmt`, `go test ./internal/api/...`
and `./internal/config/...` all clean. 108 frontend unit tests. `npm run build` clean.

**7. Schema/loader validity** (from the earlier commits, re-run): 24/24
`validate` and `validate --merged` runs across every combination, unedited plus
size-mode and offset-mode edits.

#### Not verified

**That the image finishes building.** The build reached `PreProcess` and stopped
at `failed to install host dependencies: ... sudo ...` — the documented root
requirement (`shell.go` `sudoPrefix`); this host has no passwordless sudo rule.
Everything up to the privileged step is exercised; the build itself is not, and
no image was produced.

#### Found while verifying, not fixed here

`config.mergeAdditionalFiles` (`internal/config/merge.go:357-378`) rebuilds its
slice by ranging a `map`, so `systemConfig.additionalFiles` comes out in a
random order — **the same compose request returned 4 distinct resolved YAMLs in
8 calls.** Pre-existing and unrelated to this change, but it means the Review
pane's YAML is not stable between identical composes, and any golden test over a
resolved template will be flaky. Left alone deliberately: it is in
`internal/config`, affects every build, and whether the order is semantically
meaningful (later `additionalFiles` overwrite earlier ones at copy time) needs
its own look.

image


arodage (approved):
LGTM

PR #871, opened 2026-08-31
fix(ci): make gitleaks pre-merge blocking and drop token persistence

tejaswin98:
#### Merge Checklist  

- [x] Ready to merge

#### Description 

Fixes three CI/CD security-scanner issues raised in a recent security review:

1. **Gitleaks pre-merge scan cannot block a PR on findings.** The composite action deliberately swallowed the gitleaks return code and only emitted a `::warning::`; the caller workflow never inspected the returned exit code. Net effect: a leaked secret in a PR would upload a SARIF report to the Security tab but the PR check would still be green.
2. **The gitleaks scan's baseline file is wired to the wrong composite-action input.** The workflow passed a SARIF baseline via `config_path:`, which the composite action treats as a gitleaks TOML config. As a result, any intended baseline suppression would either error out or silently do nothing.
3. **The gitleaks workflow ran `actions/checkout` with `persist-credentials: true`.** This leaves `GITHUB_TOKEN` in `.git/config` for the whole job, giving every subsequent step (including third-party or composite actions) a path to read the token. Every other workflow in this repo already uses `persist-credentials: false`; this workflow was the outlier.

Changes:

- **`.github/workflows/gitleak-scan.yml`**
  - `persist-credentials: true` → `false`; added `fetch-depth: 0` because `scan-scope: all` needs full git history.
  - Renamed the baseline input from `config_path:` to `baseline_path:` so the file is passed to the correct composite-action input.
  - Added `id: gitleaks`, an explicit `exit_code_on_leak: "1"`, and a follow-up `Fail on gitleaks findings` step that reads `steps.gitleaks.outputs.exit_code` and fails the job on non-zero (belt-and-braces gate — even if the composite action is later reverted, this step still catches findings).

- **`.github/actions/security/gitleaks/action.yml`**
  - Removed the deliberate "don't hard-fail the job" behavior. The composite action now exits with the gitleaks return code and emits `::error::` on findings. Callers wanting advisory behavior must explicitly set `exit_code_on_leak: "0"`, which is auditable in the workflow file.

Companion PR in `orch-ci` fixing the same class of bug centrally for all calling repos: [open-edge-platform/orch-ci#`fix/blocking-security-scans`](https://github.com/open-edge-platform/orch-ci/pull/new/fix/blocking-security-scans). **Independent — no merge-order dependency.**

Docs: no user-facing behavior changes; no doc update needed.

None. This PR only rewires existing scanners (gitleaks + its already-vendored composite action) to fail-closed on findings.

- **YAML syntax:** validated both changed files parse with `python3 -c "import yaml; yaml.safe_load(open('<file>'))"`.
- **Grep verification** in the repo confirms:
  - No remaining `persist-credentials: true` in this workflow.
  - `baseline_path:` is the only argument used for the baseline file.
  - The composite action's `if [[ "$STATUS" -ne 0 ]]` branch now hard-exits.
- **Backward compatibility:** the composite action's existing guard `[[ -n "$INPUT_BASELINE" && -f "$INPUT_BASELINE" ]]` tolerates a missing baseline file, so no baseline file needs to be created for this PR to function correctly.
- Build / runtime code is not exercised — this PR only touches CI security-scan wiring.

### DIFF
--- a/.github/workflows/gitleak-scan.yml
+++ b/.github/workflows/gitleak-scan.yml
@@ -22,6 +22,7 @@ jobs:
           scan-scope: "all"
           source: "./"
           version: "8.28.0"
+          config_path: ".github/gitleaks.toml"
           report_format: "sarif"
           redact: "true"
           exit_code_on_leak: "1"
--- a/.github/workflows/gitleak-scan.yml
+++ b/.github/workflows/gitleak-scan.yml
@@ -22,7 +22,6 @@ jobs:
           scan-scope: "all"
           source: "./"
           version: "8.28.0"
-          config_path: ".github/gitleaks.toml"
           report_format: "sarif"
           redact: "true"
           exit_code_on_leak: "1"
--- a/.github/workflows/gitleak-scan.yml
+++ b/.github/workflows/gitleak-scan.yml
@@ -13,13 +13,21 @@ jobs:
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 #v7.0.1
         with:
-          persist-credentials: true
+          persist-credentials: false
+          fetch-depth: 0
       - name: Run Gitleaks scan
+        id: gitleaks
         uses: ./.github/actions/security/gitleaks
         with:
           scan-scope: "all"
           source: "./"
           version: "8.28.0"
-          config_path: "./ci/gitleaks_baselines/image-composer-tool-gitleaks.sarif"
+          baseline_path: "./ci/gitleaks_baselines/image-composer-tool-gitleaks.json"
           report_format: "sarif"
           redact: "true"
+          exit_code_on_leak: "1"
+      - name: Fail on gitleaks findings
+        if: steps.gitleaks.outputs.exit_code != '0'
+        run: |
+          echo "::error::Gitleaks detected potential secrets. See SARIF in Security tab."
+          exit 1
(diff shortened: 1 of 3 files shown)


.github/workflows/gitleak-scan.yml
```diff
@@ -13,13 +13,15 @@ jobs:
...
+          persist-credentials: false
+          fetch-depth: 0
       - name: Run Gitleaks scan
+        id: gitleaks
         uses: ./.github/actions/security/gitleaks
         with:
           scan-scope: "all"
           source: "./"
           version: "8.28.0"
-          config_path: "./ci/gitleaks_baselines/image-composer-tool-gitleaks.sarif"
```
arodage, line 23: Was this config not used?
Do we need to source it from somewhere else?

tejaswin98: That line pre-dates this PR and was a no-op — config_path expects a gitleaks TOML rules config (not a SARIF baseline), and the referenced file was never committed, so the composite's file guard skipped it silently. Gitleaks was already running with defaults; removing the line makes that explicit. If we need baseline suppression later, we can add a real JSON baseline via baseline_path: in a follow-up PR.

arodage: got it

arodage (approved):
LGTM

PR #885, opened 2026-09-08
docs(adr): add EdgePack integration decision

magerstam:
#### Merge Checklist  

- [X] Ready to merge

#### Description 

Adds an ADR documenting EdgePack integration with os-image-composer.

Key decisions:

- Build EdgePack DKMS modules during image composition inside the target Ubuntu chroot.
- Use the exact target kernel and headers rather than the composer host kernel.
- Invoke DKMS explicitly with the target kernel release.
- Define Secure Boot signing, validation, and future HWE upgrade handling.
- Require validation of module presence, signatures, vermagic, DKMS status, depmod, and initramfs output.
- Document package, signing-key, and target-kernel compatibility requirements.

No related issue provided.

None. This ADR documents the use of existing Ubuntu packages and tools, including DKMS, kernel headers, kmod, compiler tooling, and EdgePack packages. It does not add any project dependencies

Documentation-only validation was performed

### DIFF
--- a/docs/architecture-decision-record/adr-edgepack-integration.md
+++ b/docs/architecture-decision-record/adr-edgepack-integration.md
@@ -546,8 +546,8 @@ behavior.
 
 ## References
 
-- EdgePack package manifest: `data/edgepacks-template.yml`
-- EdgePack APT integration: `edgepack_shared/install_logic.py`
+- EdgePack package manifest (external EdgePack repo): `edgepacks-template.yml`
+- EdgePack APT integration logic (external EdgePack repo): `edgepack_shared/install_logic.py`
 - Current Noble repository: (internal)
 - `intel-edge-base-standard` 1.0~noble14 Debian control metadata
 - `intel-edge-graphics-core` 1.0~noble11 Debian control metadata
--- a/docs/architecture-decision-record/adr-edgepack-integration.md
+++ b/docs/architecture-decision-record/adr-edgepack-integration.md
@@ -1,0 +1,558 @@
+# ADR: Integrate EdgePack into ICT OS Image Composer
+
+**Status**: Proposed
+**Date**: 2026-09-08
+**Updated**: 2026-09-08 — Integration EdgePack & Secure Boot to ICT
+**Authors**: ICT Team
+**Technical Area**: OS image composition, Ubuntu packaging, kernel modules, DKMS
+
+---
+
+## Summary
+
+Integrate EdgePack into ICT images as an APT-managed package set layered on a
+validated Ubuntu release, architecture, and kernel ABI. The base package for
+this integration is `intel-edge-base-standard`. For the initial image,
+os-image-composer will build and install all required EdgePack DKMS modules
+**during composition** for the image's **explicit target kernel**, not
+implicitly for the composer's running kernel. First boot will verify and load
+the already-built modules; it will not normally compile them.
+
+The image manifest must resolve the exact target kernel release before the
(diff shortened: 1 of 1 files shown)


docs/architecture-decision-record/adr-edgepack-integration.md
```diff
@@ -0,0 +1,558 @@
...
+corresponding certificate becomes trusted by the target platform, and whether
+the same signing identity is available for future on-target DKMS rebuilds.
+
+### Signing policy in the image manifest
+
+The image manifest should carry a signing **policy and references to key
+material**, not inline private-key contents. A representative schema is:
+
+```yaml
+edgepack:
```
yktyagi, line 257: This adds a new top-level `edgepack:` block with snake_case keys, but the rest of the schema keeps kernel/package/DKMS config under `systemConfig` (camelCase). It also ties a vendor name into what should be a generic mechanism — the next vendor with the same DKMS + Secure Boot need would end up wanting their own top-level block too. Could this extend `systemConfig.dkms` instead of adding a new namespace?

docs/architecture-decision-record/adr-edgepack-integration.md
```diff
@@ -0,0 +1,558 @@
...
+  kernel:
+    package: linux-image-<target-kernel>
+    headers: linux-headers-<target-kernel>
+    release: <target-kernel>
+  dkms:
+    build_during_composition: true
+    secure_boot:
```
yktyagi, line 264: The Error Handling section later in this doc validates that built modules are properly signed, but is there validation at manifest-parse time that `signing_private_key`/`signing_certificate` are actually present when `secure_boot.enabled: true`? Without that, a manifest could set `enabled: true` with those fields blank or missing and only fail much later during the build.

docs/architecture-decision-record/adr-edgepack-integration.md
```diff
@@ -0,0 +1,558 @@
...
+      enabled: true
+      # Supplied to the composer through a protected secret/file mechanism.
+      # Do not embed PEM/DER private-key data directly in this YAML.
+      signing_private_key: /run/secrets/edgepack/module-signing.key
+      signing_certificate: /run/secrets/edgepack/module-signing.crt
+      # Copy the signing identity into the resulting image so future
+      # target-side DKMS builds can sign modules after kernel updates.
+      retain_signing_identity: true
```
yktyagi, line 272: If this is `true`, is the same signing key embedded into every image built from this manifest, or is a unique key expected per device/build? If it's shared across a fleet, one compromised device's retained key would let an attacker sign trusted modules for every other device enrolled with the same certificate. Worth stating whether per-device key uniqueness is required.

docs/architecture-decision-record/adr-edgepack-integration.md
magerstam: Good point -- this was not meant to be normative but updated the sample section. We could move that in there

docs/architecture-decision-record/adr-edgepack-integration.md
magerstam: Yes, that pre-check should be done. Updated.

docs/architecture-decision-record/adr-edgepack-integration.md
magerstam: That's true, but I don't think there is a simple way in which we can mitigate that risk here. We should be applying right OS level authorization for this key jus tlike with any key you have on the system.

arodage (approved):
LGTM

